Issue #7, opened 2025-07-30
Add model customization


marcoabreu:
Could you add model selection as per the latest release? https://github.com/anthropics/claude-code/blob/main/CHANGELOG.md#1064

wshobson:
Good idea. I'll look into it.

wshobson:
Model customization has been implemented for all 46 subagents.

Each agent now specifies which Claude model to use based on task complexity:
- Haiku 3.5 for simple tasks (8 agents)
- Sonnet 4 for development work (26 agents)  
- Opus 4 for complex tasks (11 agents)

This ensures cost-effective AI usage while maintaining quality where it matters most. Thanks for the suggestion!

AndyBaird:
Just cloned the latest into claude/agents and all agents default to sonnet.

wshobson:
Quick update: While implementing model customization, I discovered that one of the models was incorrectly specified as 'claude-opus-4-20250514' when it should be simply 'opus' (as per the latest claude-code conventions).

This has been corrected in the prompt-engineer agent. The shortened model names appear to be the new standard, though documentation on this change is limited. All other agents may need similar updates if they follow this new naming convention.

PR #18, opened 2025-07-31
Add model customization to php-pro subagents

yenaryenar:
#7 PHP, like other languages, specifies the Sonnet 4 model.


wshobson:
I updated it with the proper convention @yenaryenar . Thanks!

PR #46, opened 2025-08-08
Add ruby-pro agent

ayaman:
Add ruby-pro agent and update README.


wshobson (approved):
Thanks @ayaman . Look great!

PR #118, opened 2025-11-14
Replace GPT and Claude models to latest, better and cheaper models

kunalshah:
Use Claude Sonnet 4.5 instead of 3.5 and GPT 5 instead of GPT 4.


wshobson (approved):
Thanks for keeping these up-to-date, @kunalshah.

PR #128, opened 2025-11-21
docs(agents): add haskell-pro

jsjoeio:
Noticed there wasn't one for Haskell so thought I should add one


wshobson (dismissed):
Thanks for the contribution, @jsjoeio\! A Haskell agent is a great addition - the agent itself is well-written with comprehensive coverage of type systems, STM concurrency, and the Haskell ecosystem.

However, there's an issue with the file location that needs to be addressed before we can merge:

### Issue: Non-existent plugin directory

The PR creates the agent at `plugins/haskell-development/agents/haskell-pro.md`, but the `haskell-development` plugin doesn't exist. This would leave the agent file orphaned.

### Requested change

Please move the agent to the existing `functional-programming` plugin:

```
plugins/functional-programming/agents/haskell-pro.md
```

This is where `elixir-pro` lives, and it makes sense to group functional languages together.

You'll also need to update the path in `docs/agents.md`:

```markdown
| [haskell-pro](../plugins/functional-programming/agents/haskell-pro.md) | sonnet | Strongly typed functional programming with purity, advanced type systems, and concurrency |
```

Once you make these changes, this will be good to merge. Thanks again for adding Haskell support\!

PR #435, opened 2026-02-19
Recommend modern Stripe best practices

sawyerh:
Hi there 👋🏻 I work at Stripe.

## Changes

- Clarify that Checkout Sessions is the recommended path for not just Hosted Checkout, but [also Elements](https://docs.stripe.com/payments/checkout-sessions-and-payment-intents-comparison). 
- Remove Card Element example. [Card Element is a legacy element, and Payment Element is the suggested replacement](https://docs.stripe.com/payments/payment-card-element-comparison).
- Remove the hard-coded `payment_method_types` param in Checkout Session creation requests. [Dynamic payment methods](https://docs.stripe.com/payments/payment-methods/dynamic-payment-methods) are recommended.

### DIFF
--- a/plugins/payment-processing/skills/stripe-integration/SKILL.md
+++ b/plugins/payment-processing/skills/stripe-integration/SKILL.md
@@ -214,6 +214,9 @@ def create_payment_intent(amount, currency='usd', customer_id=None):
         automatic_payment_methods={
             'enabled': True,
         },
+        metadata={
+            'integration_check': 'accept_a_payment'
+        }
     )
     return intent.client_secret  # Send to frontend
 ```
@@ -517,7 +520,3 @@ def test_payment_flow():
 - **Hardcoded Amounts**: Use cents/smallest currency unit
 - **No Retry Logic**: Implement retries for API calls
 - **Ignoring Test Mode**: Test all edge cases with test cards
-
-```
-
-```


wshobson:
Hey @sawyerh — thanks for this PR! Great to have guidance directly from someone at Stripe. The direction is spot on and aligns with current Stripe docs. I've reviewed everything against the latest documentation and have two sets of feedback:

## Small fixes I'd like to apply

These are minor polish items I can push as a follow-up commit (or you're welcome to incorporate):

1. **Quick Start section (line ~80)** still has `payment_method_types=['card']` — should be removed for consistency with the dynamic payment methods recommendation you're making in Pattern 1
2. **Testing section (line ~435)** also still has `payment_method_types=['card']` — should use `automatic_payment_methods={'enabled': True}` instead
3. **Unused `appearance` variable** in the Pattern 2 JS example — declared but never referenced. Should be removed or wired up
4. **`actions` access inconsistency** — `loadActionsResult.actions.getSession()` is called before destructuring `const {actions} = loadActionsResult;`. Cleaner to destructure first, then use `actions.getSession()`
5. **Embedded checkout bullet** — the Custom UI bullet shows `ui_mode='custom'` but the Embedded bullet doesn't show `ui_mode='embedded'`. Adding it would be consistent
6. **Clarifying comments** — a note that `initCheckout()` is synchronous while `loadActions()` is async, and a note on `session.total.total.amount` explaining it's the grand total in smallest currency unit

## Suggested addition: modernized Payment Intents pattern

The PR removes the old Payment Intents + Card Element pattern entirely (rightfully so — Card Element is legacy). But I think developers who genuinely need bespoke control should still have a reference pattern. I've drafted a **Pattern 3: Payment Intents with Payment Element (Bespoke Control)** that:

- Uses **Payment Element** instead of the removed Card Element
- Uses `stripe.confirmPayment()` instead of `stripe.confirmCardPayment()`
- Uses `automatic_payment_methods` instead of hardcoded `payment_method_types`
- Splits Python and JavaScript into separate fenced code blocks (cleaner for a markdown file)
- Includes clear guidance: *"Use this when you need full control and cannot use Checkout Sessions"*

Here's the pattern I'd add between your Pattern 2 and the Subscription pattern:

```python
def create_payment_intent(amount, currency='usd', customer_id=None):
    """Create a payment intent for bespoke checkout UI with Payment Element."""
    intent = stripe.PaymentIntent.create(
        amount=amount,
        currency=currency,
        customer=customer_id,
        automatic_payment_methods={
            'enabled': True,
        },
    )
    return intent.client_secret  # Send to frontend
```

```javascript
// Frontend: Mount Payment Element and confirm via Payment Intents
const stripe = Stripe('pk_test_...');
const elements = stripe.elements({clientSecret});

const paymentElement = elements.create('payment');
paymentElement.mount('#payment-element');

document.getElementById('pay-button').addEventListener('click', async () => {
    const {error} = await stripe.confirmPayment({
        elements,
        confirmParams: {
            return_url: 'https://yourdomain.com/complete',
        },
    });

    if (error) {
        document.getElementById('errors').textContent = error.message;
    }
});
```

Does this look right to you? Happy to push these changes or let you incorporate them — whatever you prefer!

wshobson (approved):
Thanks so much for this, Sawyer — really appreciate you taking the time to modernize the patterns and then circling back to address all the feedback. Having guidance directly from someone at Stripe makes a huge difference. Merging this up!

PR #478, opened 2026-04-10
fix: monte carlo layer uses nonexistent sdk.stream()

### LINKED ISSUE #477: [BUG] Monte Carlo layer uses nonexistent sdk.stream() instead of sdk.query()
### Preliminary Checks

- [x] I have read the [Code of Conduct](https://github.com/wshobson/agents/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] I have searched existing issues to ensure this is not a duplicate

### Affected Subagent

plugin-eval (Layer 3: monte_carlo)

### Bug Description

`run_simulation()` in `src/plugin_eval/layers/monte_carlo.py` calls `sdk.stream()`, which does not exist in `claude-agent-sdk`. The judge layer correctly uses `sdk.query()` with `ClaudeAgentOptions` and `ResultMessage`, but the monte carlo layer references a nonexistent API.

Because the call is wrapped in a bare `except`, every simulation silently fails with `errored=True`. This causes Layer 3 to report 100% failure rate, which tanks the composite score on dimensions like robustness, triggering accuracy, and token efficiency.

### Steps to Reproduce

1. Install with LLM extras: `uv run --extra llm plugin-eval certify <any-skill-path>`
2. Observe the monte_carlo layer score is artificially low
3. Add logging in the `except` block to confirm: `AttributeError: module 'claude_agent_sdk' has no attribute 'stream'`

### Expected Behavior

Monte Carlo simulations should run successfully using `sdk.query()`, the same pattern the judge layer uses in `layers/judge.py`.

### Additional Context

The fix is replacing the `sdk.stream()` block in `run_simulation()` with `query()` + `ClaudeAgentOptions` + `ResultMessage`, matching the working pattern in `judge.py`. PR at https://github.com/wshobson/agents/pull/478

slgoodrich:
## Summary

- `run_simulation()` in `monte_carlo.py` calls `sdk.stream()`, which doesn't exist in `claude-agent-sdk`
- Every MC simulation silently errors (caught by bare `except`), causing 100% failure rate in Layer 3
- Replaced with the `query()`/`ClaudeAgentOptions`/`ResultMessage` pattern already used by the judge layer

## Test results (prd-templates skill)

| Metric | Before | After |
|--------|--------|-------|
| monte_carlo score | 0.300 | 0.600 |
| Robustness | 0.000 (F) | 1.000 (A+) |
| Token Efficiency | 0.439 (F) | 0.995 (A+) |
| Overall | 42.0/100 | 50.5/100 |

Fixes #477

### DIFF
--- a/plugins/plugin-eval/src/plugin_eval/layers/monte_carlo.py
+++ b/plugins/plugin-eval/src/plugin_eval/layers/monte_carlo.py
@@ -53,7 +53,11 @@ class MonteCarloConfig:
 async def run_simulation(skill_content: str, prompt: str, auth: str) -> SimResult:
     """Run a single simulation via Agent SDK. Returns SimResult. On error, errored=True."""
     try:
-        import claude_agent_sdk as sdk  # type: ignore[import-untyped]
+        from claude_agent_sdk import (  # type: ignore[import-untyped]
+            ClaudeAgentOptions,
+            ResultMessage,
+            query,
+        )
 
         result_text = ""
         activated = False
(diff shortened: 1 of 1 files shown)


slgoodrich-acf:
@wshobson whoops. Was doing some repo cleanup and forgot about this PR. Can redo it if you want, but the fix is pretty straight forward.

PR #484, opened 2026-04-15
feat: add protect-mcp plugin — Cedar policies + signed receipts (closes #471)

tomjwxf:
## Summary

Closes #471.

Adds `protect-mcp`, the first cryptographic governance plugin in the marketplace. Every Claude Code tool call is:

1. **Evaluated** against a Cedar policy before it runs (PreToolUse hook). Cedar `deny` → exit 2 → tool blocked.
2. **Signed** with an Ed25519 key after it runs (PostToolUse hook). Receipts are hash-chained and written to `./receipts/`.
3. **Verifiable offline** by anyone via `npx @veritasacta/verify` — no network, no vendor lookup, no trust in the operator.

Directory structure matches what was requested in #471:

```
plugins/protect-mcp/
├── .claude-plugin/plugin.json
├── README.md
├── skills/protect-mcp-setup/SKILL.md
├── agents/policy-enforcer.md        (opus)
├── agents/receipt-verifier.md       (sonnet)
├── commands/verify-receipt.md       (/verify-receipt <path>)
├── commands/audit-chain.md          (/audit-chain [--last N])
└── hooks/hooks.json
```

Also adds the marketplace entry under `category: security`. If you'd prefer a new `governance` category, happy to add one in a follow-up — I defaulted to `security` since that category already exists.

## Why this is a good fit

- **Addresses an open issue.** #471 specifically requested this plugin with this structure.
- **Depends on published packages.** The plugin is a thin wrapper around two npm packages that already exist:
  - [`protect-mcp`](https://www.npmjs.com/package/protect-mcp) — 10K+ monthly downloads, the hooks runtime
  - [`@veritasacta/verify`](https://www.npmjs.com/package/@veritasacta/verify) — offline verification CLI
- **Standards-based.** Ed25519 (RFC 8032), JCS (RFC 8785), Cedar (AWS's open authorization engine), IETF draft [draft-farley-acta-signed-receipts](https://datatracker.ietf.org/doc/draft-farley-acta-signed-receipts/).
- **Ecosystem presence already exists.** Cedar WASM bindings for agent policies merged in [cedar-policy/cedar-for-agents#64](https://github.com/cedar-policy/cedar-for-agents/pull/64) and a governed example merged in [microsoft/agent-governance-toolkit#667](https://github.com/microsoft/agent-governance-toolkit/pull/667).

## Test plan

- [x] `python3 -m json.tool` validates `plugin.json`, `hooks.json`, and the updated `marketplace.json`
[...]

PR #494, opened 2026-04-17
test(protect-mcp): add test/ fixtures and round-trip verification

tomjwxf:
Follow-up to #484 closing the test-plan commitment made in that PR. Adds a
`plugins/protect-mcp/test/` directory with deterministic fixtures that
exercise the full evaluate → sign → verify loop, including the tamper
detection path.

## What's inside

| File | Purpose |
|------|---------|
| `fixtures/test-policy.cedar` | Cedar policy exercising both `permit` and `forbid` semantics |
| `fixtures/pretool-allow-read.json` | Input that should be permitted |
| `fixtures/pretool-allow-bash-safe.json` | `git status` — permitted |
| `fixtures/pretool-deny-bash-destructive.json` | `rm -rf /` — denied |
| `fixtures/pretool-deny-write.json` | Unscoped `Write` — denied |
| `fixtures/posttool-signing-input.json` | Input for receipt signing |
| `expected/receipt-schema.json` | JSON Schema pinned to draft-farley-acta-signed-receipts required fields |
| `run-tests.sh` | Full round-trip (8 tests, requires node >= 18 + python3) |
| `verify-fixtures.sh` | Static validation, python3-only, safe for sandboxed CI |
| `README.md` | Layout, running instructions, extension guide |

## Test matrix (run-tests.sh)

| # | Scenario | Expected exit |
|---|----------|----------------|
| 1 | PreToolUse on `Read` | 0 (permit) |
| 2 | PreToolUse on `Bash git status` | 0 (permit) |
| 3 | PreToolUse on `Bash rm -rf /` | 2 (forbid) |
| 4 | PreToolUse on `Write` | 2 (forbid) |
| 5 | PostToolUse signing produces a receipt file | 0 |
| 6 | Produced receipt conforms to the schema | 0 |
| 7 | `@veritasacta/verify` accepts the receipt | 0 |
| 8 | Tampered receipt is rejected | 1 |

**Test 8 is the critical regression guard.** Flipping the `decision` field
in a signed receipt must invalidate the Ed25519 signature, so
`@veritasacta/verify` must exit 1 rather than 0. This locks in the tamper
detection property that the plugin claims.

## Why two scripts

`run-tests.sh` needs `npx` and fetches `protect-mcp` + `@veritasacta/verify`
from the npm registry on first run. That is fine for local dev and well-
provisioned CI, but not always available in sandboxed checks.

`verify-fixtures.sh` is `python3`-only and purely local: validates JSON shape,
required fields, and cedar non-emptiness. It is the safe static gate that can
run anywhere. Both scripts use exit code 77 for "required tool missing," the
autotools convention that most CI frameworks interpret as "skipped."

## What I verified

- `bash plugins/protect-mcp/test/verify-fixtures.sh` passes locally (12/12)
- `run-tests.sh` structure mirrors what the existing hook commands accept
  (same `--policy`, `--tool`, `--input`, `--output`, `--receipts`, `--key`
  flags as `hooks/hooks.json`)
- Every fixture is valid JSON against Python's `json.load`
- `expected/receipt-schema.json` is valid JSON Schema draft-07 and requires
  all fields emitted by `protect-mcp@latest sign`

## Notes

- No changes to the plugin itself, no changes to `marketplace.json` or
  `hooks.json`, no new runtime dependencies
- Scripts are marked executable (`chmod +x`)
- The Cedar policy in `fixtures/` is narrower than the production example
  in `skills/protect-mcp-setup/SKILL.md`; it is optimized for test
  determinism rather than real-world coverage

Closes the `test plan` checklist from #484. Happy to iterate on scope or
rename/restructure to match any existing test conventions in the marketplace.

### DIFF
--- a/plugins/protect-mcp/test/README.md
+++ b/plugins/protect-mcp/test/README.md
@@ -1,0 +1,94 @@
+# protect-mcp test fixtures
+
+Round-trip tests for the `protect-mcp` plugin's `PreToolUse` and `PostToolUse`
+hooks. Exercises the full evaluate → sign → verify loop against deterministic
+fixtures, including the tamper-detection path.
+
+## Layout
+
+```
+test/
+├── fixtures/
+│   ├── test-policy.cedar                  # Cedar policy used by all tests
+│   ├── pretool-allow-read.json            # Read should be permitted
+│   ├── pretool-allow-bash-safe.json       # Bash "git status" should be permitted
+│   ├── pretool-deny-bash-destructive.json # Bash "rm -rf /" should be denied
+│   ├── pretool-deny-write.json            # Write should be denied
+│   └── posttool-signing-input.json        # Input for receipt signing
+├── expected/
+│   └── receipt-schema.json                # Expected receipt shape (JSON Schema)
+├── run-tests.sh                           # Full round-trip (requires node / npx)
+└── verify-fixtures.sh                     # Static validation (python3 only)
+```
+
+## Running
+
+### Full round-trip (local development)
+
+```bash
+./run-tests.sh
+```
+
+Requires `node` (>= 18), `npx`, and `python3`. Fetches `protect-mcp` and
+`@veritasacta/verify` from npm on first run. Runs eight tests:
+
+| # | Scenario | Expected exit |
+|---|----------|----------------|
+| 1 | `PreToolUse` on `Read`                         | 0 (permit)  |
+| 2 | `PreToolUse` on `Bash git status`              | 0 (permit)  |
+| 3 | `PreToolUse` on `Bash rm -rf /`                | 2 (forbid)  |
+| 4 | `PreToolUse` on `Write`                        | 2 (forbid)  |
+| 5 | `PostToolUse` signing produces a receipt file  | 0 (success) |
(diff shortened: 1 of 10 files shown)


wshobson:
Thanks @tomjwxf — merged. Nice two-tier design: `verify-fixtures.sh` as the fast static gate and `run-tests.sh` for the full round-trip, with autotools exit-77 for skip. Test 8 (tamper detection) is exactly the regression guard this plugin needs.

One optional follow-up: `run-tests.sh` uses `npx protect-mcp@latest`, so an upstream release could silently turn the tamper-detection guard green or red without any repo change. Pinning to a specific `protect-mcp@x.y.z` (and bumping deliberately) would make the guard fully reproducible. Not blocking — happy to land as-is.

tomjwxf:
Thanks for the merge @wshobson and for the clean review across this batch. The pinning suggestion is a real point — `npx protect-mcp@latest` in test 8 means an upstream release could swing the tamper-detection guard without any repo-side signal. I will open a small follow-up PR pinning `protect-mcp@<specific-version>` across all three governance plugins (this test, plus the runtime hooks in protect-mcp and review-agent-governance) once the two open ones (#495, #496) rebase cleanly. Deliberate bumps only from then on.

PR #490, opened 2026-04-17
fix: add missing YAML frontmatter to codebase-cleanup commands

xiaolai:
> **Automated audit**: This PR was generated by [NLPM](https://github.com/xiaolai/nlpm-for-claude), a natural language programming linter, running via [claude-code-action](https://github.com/anthropics/claude-code-action). Please evaluate the diff on its merits.

## Bug

All three commands in `plugins/codebase-cleanup/commands/` have no YAML frontmatter block:

- `tech-debt.md` — starts with `# Technical Debt Analysis and Remediation`
- `deps-audit.md` — starts with `# Dependency Audit and Security Analysis`
- `refactor-clean.md` — starts with `# Refactor and Clean Code`

Claude Code requires a `---` delimited frontmatter block with at minimum a `description` field to register a slash command. Without it, the command is silently absent when users install the `codebase-cleanup` plugin — all three slash commands are completely unavailable.

## Fix

Added minimal frontmatter to each command with a `description` derived from the file's title and opening paragraph:

```yaml
---
description: "..."
---
```

No content was changed — only the required registration header was prepended. The descriptions are concise, single-sentence summaries of what each command does.

PR #495, opened 2026-04-17
feat: add review-agent-governance plugin (second inhabitant of governance category)

tomjwxf:
Hi @wshobson, follow-up contribution to the `governance` category you
created when merging #484. This one targets a specific class of failure:
AI agents posting review-surface actions (PR reviews, comments, merges,
CI edits) without a human gate.

## Context for why this plugin

Review bots posting hallucinated review comments under real accounts is
a pattern common enough to name. The damage is immediate and visible:
reviews show up under the account, merges happen that should not, CI
workflow files get rewritten. Most of these incidents would have been
prevented by a simple pattern: **require a human-signal file to exist
before the agent is allowed to take review-surface actions.**

This plugin implements that pattern with Cedar + Ed25519 receipts, so
the audit trail is cryptographic and the policy is declarative.

## What's in the PR

```
plugins/review-agent-governance/
├── .claude-plugin/plugin.json
├── README.md                                       # motivation + full usage
├── skills/review-agent-setup/SKILL.md              # step-by-step setup
├── agents/review-policy-author.md      (sonnet)    # Cedar author for review policies
├── commands/
│   ├── approve-review.md                           # /approve-review "<reason>"
│   └── list-pending.md                             # /list-pending [--last N]
├── hooks/hooks.json                                # PreToolUse gate + PostToolUse sign
└── policies/review-agent-governance.cedar          # default policy (5 forbid rules)
```

Plus one entry added to `.claude-plugin/marketplace.json` under
`category: "governance"` (the category you created in #484). This turns
the governance category from a vanity shelf into a real one with two
complementary plugins.

## What it forbids by default

The Cedar policy forbids (unless `./.review-approved` is present):

- `gh pr review|comment|merge|close|edit`, `gh issue comment|close|edit`,
  `gh release create|edit`, `gh api repos`
- GitLab / Bitbucket equivalents (`glab mr ...`)
- `git push` to `main`, `master`, `release`, `production`
- Writes to `.github/workflows/`, `.github/CODEOWNERS`, `.gitlab-ci.yml`,
  `.circleci/config.yml`
- `WebFetch` POSTs to `api.github.com`, `api.gitlab.com`, `hooks.slack.com`,
  `discord.com`

Non-review actions pass through unchanged. Installs cleanly alongside
`protect-mcp` (different receipt directory: `./review-receipts/` vs `./receipts/`).

## The approval flow

**Flag file:**
```bash
touch ./.review-approved
# agent makes the approved action
rm ./.review-approved
```

**Slash command:**
```
/approve-review "Approving LGTM on PR #42"
```
Creates the flag with the reason embedded, logs an approval receipt.

**List denials:**
```
/list-pending
```
Walks the receipt chain and shows recent `decision: deny` entries so
the human can see what the agent tried that was blocked.

Every action (approved or denied) produces an Ed25519-signed receipt.
The chain is verifiable offline via `npx @veritasacta/verify ./review-receipts/*.json`.

## Why this is the right second inhabitant for the category

1. **Different class than protect-mcp.** `protect-mcp` is general-purpose:
   any tool call, any policy. This plugin is focused: review-surface
   actions, with a known-good default policy and an explicit human-gate
   pattern. Users who need both install both; users who only need
   review gating install just this.

2. **Solves a common real pain.** The review-bot failure class is
   something many teams have felt. A plugin that makes the fix one-line
   is genuinely useful.

3. **Cements the `governance` shelf.** One plugin in a category reads as
   a vanity placement. Two that solve different problems reads as a
   real category.

## Verification

- [x] `python3 -m json.tool` validates `plugin.json`, `hooks.json`, and
      the updated `marketplace.json`
- [x] Cedar policy is well-formed (9 rules; 4 forbid + 1 permit)
- [x] README and SKILL document the approval flow end-to-end
- [x] Slash commands are self-contained and runnable via Bash
- [x] Marketplace entry uses the same shape and naming convention as
      `protect-mcp` (category `governance`, keywords array, URL author)

## Notes

- The plugin depends on the same `protect-mcp` npm package as PR #484,
  so users who installed that one already have the runtime. No new
  dependencies.
- Default receipts directory is `./review-receipts/` rather than
  `./receipts/` to avoid collision with protect-mcp when both are
  installed. Users can override via `REVIEW_GOVERNANCE_RECEIPTS`.
- The file-flag approval pattern is deliberately simple for v0.1. A
  future v0.2 could add scoped approvals (e.g., "approved for next N
  review actions only"), but v0.1 is useful as-is.

Happy to iterate on scope, naming, or the default policy coverage.
Thanks for the governance category in #484 — made this contribution
possible as a proper shelf-mate rather than a category-of-one.

cc @wshobson

### DIFF
--- a/plugins/review-agent-governance/policies/review-agent-governance.cedar
+++ b/plugins/review-agent-governance/policies/review-agent-governance.cedar
@@ -1,0 +1,103 @@
+// review-agent-governance: default Cedar policy
+//
+// Forbids an AI agent from taking consequential review actions without an
+// explicit human approval signal. A human opens a short approval window by
+// creating a ./.review-approved flag file (or using the /approve-review
+// slash command shipped with this plugin). When the flag is present, the
+// PreToolUse hook short-circuits to permit; when it is absent, this policy
+// evaluates.
+//
+// Every tool call, approved or not, still produces an Ed25519 receipt via
+// the PostToolUse hook. The chain records what was approved and when.
+
+// ─────────────────────────────────────────────────────────────────────────
+// Forbid rules: the set of actions that require human approval
+// ─────────────────────────────────────────────────────────────────────────
+
+// 1. Posting review-surface GitHub actions via gh CLI
+forbid (
+    principal,
+    action == Action::"Bash",
+    resource
+) when {
+    context.command_pattern in [
+        "gh pr review",
+        "gh pr comment",
+        "gh pr close",
+        "gh pr merge",
+        "gh pr edit",
+        "gh issue comment",
+        "gh issue close",
+        "gh issue edit",
+        "gh release create",
+        "gh release edit",
+        "gh api repos"
+    ]
+};
+
+// 2. Posting via GitLab / Bitbucket CLIs
+forbid (
+    principal,
+    action == Action::"Bash",
+    resource
+) when {
+    context.command_pattern in [
+        "glab mr comment",
+        "glab mr approve",
+        "glab mr merge",
+        "glab issue comment"
+    ]
+};
+
+// 3. Git operations on protected branches
+forbid (
+    principal,
+    action == Action::"Bash",
+    resource
+) when {
+    context.command_pattern in ["git push", "git push --force", "git push -f"] &&
+    context.target_branch in ["main", "master", "release", "production"]
+};
+
+// 4. Direct writes to CI / CD configuration paths. Modifying these files lets
+//    an agent change how future builds or reviews run, bypassing every other
+//    safeguard. Always require human approval.
+forbid (
+    principal,
+    action in [Action::"Write", Action::"Edit"],
+    resource
+) when {
+    context.path_starts_with in [
+        ".github/workflows/",
+        ".github/CODEOWNERS",
+        ".gitlab-ci.yml",
+        ".circleci/config.yml",
+        "buildkite/pipeline.yml"
+    ]
+};
+
+// 5. WebFetch POSTs to known review / chat platforms
+forbid (
+    principal,
+    action == Action::"WebFetch",
+    resource
+) when {
+    context.method == "POST" &&
+    context.url_host in [
+        "api.github.com",
+        "api.gitlab.com",
+        "api.bitbucket.org",
+        "hooks.slack.com",
+        "discord.com"
+    ]
+};
+
+// ─────────────────────────────────────────────────────────────────────────
+// Permit everything else
+//
+// This plugin is focused. Non-review actions pass through this policy
+// unchanged. Run it alongside protect-mcp if you want general tool-call
+// policy enforcement too.
+// ─────────────────────────────────────────────────────────────────────────
+
+permit (principal, action, resource);
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -13,7 +13,7 @@
     {
       "name": "documentation-standards",
       "source": "./plugins/documentation-standards",
-      "description": "HADS (Human-AI Document Standard) — semantic tagging convention for writing docs that work efficiently for both humans and AI models. Reduces token consumption by separating machine-critical facts from human context.",
+      "description": "HADS (Human-AI Document Standard) \u2014 semantic tagging convention for writing docs that work efficiently for both humans and AI models. Reduces token consumption by separating machine-critical facts from human context.",
       "version": "1.0.0",
       "author": {
         "name": "Niksa Barlovic",
@@ -900,7 +900,7 @@
       "description": "Binary reverse engineering, malware analysis, firmware security, and software protection research for authorized security research, CTF competitions, and defensive security",
       "version": "1.0.0",
       "author": {
-        "name": "Dávid Balatoni",
+        "name": "D\u00e1vid Balatoni",
         "url": "https://github.com/balcsida"
       },
       "homepage": "https://github.com/wshobson/agents",
@@ -909,7 +909,7 @@
     },
     {
       "name": "conductor",
-      "description": "Context-Driven Development plugin that transforms Claude Code into a project management tool with structured workflow: Context → Spec & Plan → Implement",
+      "description": "Context-Driven Development plugin that transforms Claude Code into a project management tool with structured workflow: Context \u2192 Spec & Plan \u2192 Implement",
       "version": "1.2.1",
       "author": {
         "name": "Seth Hobson",
@@ -1002,7 +1002,7 @@
         "url": "https://github.com/Anasss/qa-orchestra.git",
         "path": "."
       },
-      "description": "Multi-agent QA toolkit with 10 specialized agents covering the full QA lifecycle — orchestrator, environment-manager, functional-reviewer, test-scenario-designer, browser-validator, automation-writer, manual-validator, bug-reporter, release-analyzer, and smart-test-selector. Stack-agnostic, output-chained, designed around live validation via Chrome MCP.",
+      "description": "Multi-agent QA toolkit with 10 specialized agents covering the full QA lifecycle \u2014 orchestrator, environment-manager, functional-reviewer, test-scenario-designer, browser-validator, automation-writer, manual-validator, bug-reporter, release-analyzer, and smart-test-selector. Stack-agnostic, output-chained, designed around live validation via Chrome MCP.",
       "version": "1.0.0",
       "author": {
         "name": "Anass Rach",
@@ -1015,7 +1015,7 @@
     {
       "name": "protect-mcp",
       "source": "./plugins/protect-mcp",
-      "description": "Cedar policy enforcement + Ed25519 signed receipts for every Claude Code tool call. First cryptographic governance plugin — decisions are policy-gated before they run and every decision produces a tamper-evident receipt verifiable offline.",
+      "description": "Cedar policy enforcement + Ed25519 signed receipts for every Claude Code tool call. First cryptographic governance plugin \u2014 decisions are policy-gated before they run and every decision produces a tamper-evident receipt verifiable offline.",
       "version": "0.1.0",
       "author": {
         "name": "Tom Farley",
@@ -1041,6 +1041,21 @@
       "license": "MIT",
       "category": "governance",
       "keywords": ["tutorial", "skill", "recipe", "audit", "governance", "cedar", "receipts", "ed25519"]
+    },
+    {
+      "name": "review-agent-governance",
+      "source": "./plugins/review-agent-governance",
+      "description": "Require a human approval signal before an AI agent can post PR reviews, comments, merges, or writes to CI configuration. Joins protect-mcp and signed-audit-trails in the governance category; composes with protect-mcp for runtime enforcement.",
+      "version": "0.1.0",
+      "author": {
+        "name": "Tom Farley",
+        "email": "tommy@scopeblind.com",
+        "url": "https://github.com/tomjwxf"
+      },
+      "homepage": "https://veritasacta.com",
+      "license": "MIT",
+      "category": "governance",
+      "keywords": ["review", "governance", "cedar", "receipts", "human-approval", "pr-review", "ci-guard"]
     }
   ]
 }
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -13,7 +13,7 @@
     {
       "name": "documentation-standards",
       "source": "./plugins/documentation-standards",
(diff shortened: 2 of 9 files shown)


wshobson (changes requested):
Thanks for the contribution @tomjwxf — the problem framing (hallucinated review-bot actions) is real and the Cedar coverage of the review surface is thorough. A few things to address before this can land:

**1. `marketplace.json` unicode regression (blocking).** The diff rewrites literal UTF-8 characters (`—`, `→`, `á`) as ASCII escapes (`\u2014`, `\u2192`, `\u00e1`) on *other, unrelated* plugin entries (HADS, conductor, qa-orchestra, reverse-engineering, protect-mcp). The current file uses literal UTF-8 — please only append the new `review-agent-governance` entry and leave the surrounding descriptions untouched. This is almost certainly from a `json.dump(...)` round-trip; `ensure_ascii=False` fixes it, or just append the entry by hand.

**2. `approve-review` argument handling (likely bug).** `commands/approve-review.md` uses `REASON="${1:-}"`, but slash-command bodies in this repo use `$ARGUMENTS` (see e.g. `plugins/codebase-cleanup/commands/deps-audit.md` after #490). With `$1`, `/approve-review "long reason with spaces"` captures only the first word.

**3. Approval records are unsigned.** The README at line 94 claims "The chain records exactly which actions were human-gated and when," but the approval entries under `./review-receipts/approvals/*.json` are plain JSON — they don't flow through `protect-mcp sign`, so `@veritasacta/verify` won't cover them. Either sign the approval records too, or explicitly clarify in the README that the signed chain only covers tool-call receipts and the approval log is trust-in-operator.

**Non-blocking:**

- Document that when the approval flag is present, PreToolUse short-circuits and no Cedar/`policy_digest` field is attached to the resulting PostToolUse receipt. Auditors should expect that.
- `$REASON` is embedded in a JSON body without escaping; a reason containing `"` or `\` will break the JSON. Low severity.
- Keywords-array reformatting for `protect-mcp` (single-line → multi-line) is cosmetic churn — please revert.

Also note that #496 touches the same `marketplace.json` and will conflict. Whichever you want to land first is fine, but they'll need to be rebased.

Happy to re-review once the blockers are addressed.

tomjwxf:
Thanks for the close read @wshobson — all three blockers land, and the non-blocking points were right too. Pushed 91c1b20 addressing them:

**1. marketplace.json unicode regression — fixed.** You were exactly right on the root cause (`json.dump` without `ensure_ascii=False`). Rather than rely on that flag, I reset `marketplace.json` to upstream HEAD and did a string-based insertion that only appends the new `review-agent-governance` entry. No other entries touched. Verified with `grep -c "\\u" marketplace.json` = 0. The `protect-mcp` keywords-array reformatting is also reverted as part of the reset.

**2. `approve-review` arguments — fixed.** Switched `REASON="${1:-}"` to `REASON="$ARGUMENTS"` to match the marketplace slash-command convention (saw the deps-audit example from #490). Also added a JSON-escaping pass via `python3 -c "json.dumps(...)"` before embedding the reason in the approval-record body, which resolves the non-blocking note about `"` / `\` breaking JSON.

**3. Approval records are unsigned — clarified in README.** You were right that the previous wording overstated the cryptographic coverage. Rewrote the relevant paragraph to explicitly say:

- Entries under `./review-receipts/approvals/*.json` are plain JSON, not signed. They do not flow through `protect-mcp sign`. The approval log is operator-trust.
- The signed tamper-evident chain is the `PostToolUse` tool-call receipts under `./review-receipts/*.json`. Those are what `@veritasacta/verify` covers.
- For regulated environments that need signed approval records too, pointed users at calling `npx protect-mcp@latest sign` directly.

This is the honest framing. I considered signing approval records as part of this fix but that changes the semantic model of what the approval command does (right now it is a session marker; signing it would make it a receipt in its own right), and that feels like a v0.2 design question rather than a v0.1 fix.

**4. Approval-short-circuit documentation — added.** New paragraph in the README explicitly notes that when the approval flag is present, `PreToolUse` exits 0 without running Cedar, so the downstream `PostToolUse` receipt has `decision: allow` but no `policy_digest` field. Auditors should expect that.

**On the #496 conflict:** yes, both PRs touch the same `marketplace.json` section. Happy to land whichever you prefer first and rebase the other; with the string-based insertion approach now, rebasing is mechanical.

**On version pinning (raised on #494):** good suggestion; I will open a small follow-up PR that pins `protect-mcp@<specific-version>` across all three governance plugins so the tamper-detection guard is reproducible. Will do that after these two are rebased cleanly.

Re-requesting review.

wshobson:
Thanks for this — the plugin is in good shape overall. Before merging, two items to address:

**1. Pin `protect-mcp` in `hooks/hooks.json` (blocking)**

Both hook commands use `npx protect-mcp@latest`, which reintroduces the exact reproducibility gap that #497 just closed for the `protect-mcp` plugin itself. Please pin to `protect-mcp@0.5.5` (matching #497) in both the PreToolUse and PostToolUse hooks in `plugins/review-agent-governance/hooks/hooks.json`. Also worth pinning the `@veritasacta/verify` references in README.md / SKILL.md that appear in executed example snippets — at your discretion on which are documentation-pattern vs. runtime.

**2. Rebase on main (blocking)**

Now that #496 merged, this PR conflicts at the `marketplace.json` insertion point. Please rebase and resolve — same string-append approach you used on the #495 fix commit should work.

**Non-blocking observation**

The `governance` category is now populated entirely by your plugins (protect-mcp, signed-audit-trails, and this one). That's fine given the category was explicitly created for this class of tooling, but worth naming: if future contributors want to add governance-adjacent plugins that don't use the `protect-mcp` / Cedar / Ed25519 stack, we'll want the category description to remain welcoming to those. No action needed from you on this PR.

Happy to merge once the pin lands and the rebase is clean.

tomjwxf:
@wshobson no worries, both blockers addressed in 9da068d (force-pushed).

1. protect-mcp@0.5.5 pinned in both hooks. Matching #497's pin.
2. Rebased on current main. Clean merge, no conflicts at marketplace.json
   since the string-append approach preserves the rest of the file.

The README and SKILL.md references to @veritasacta/verify stay unversioned where they document the CLI as a general pattern ("run npx @veritasacta/verify against your receipts"), since those are documentation-pattern rather than runtime invocations. If you'd prefer them pinned too, happy to follow up.

On the non-blocking observation about governance category, agreed 100%. If future contributors add governance plugins that choose something other than the protect-mcp / Cedar / Ed25519 stack, the category description at least does not prescribe a specific stack. I canrevise the marketplace.json category description to make that welcomeness explicit if a non-ScopeBlind governance submission lands. 

Follow-up PR if useful - Re-requesting review. Thanks for feedback!

tomjwxf:
Rebased onto current main, marketplace.json conflict with #496 resolved. Kept the signed-audit-trails entry and added review-agent-governance after it with single-line keywords format to match repo convention. Tweaked the description to "Joins protect-mcp and signed-audit-trails in the governance category" rather than the original "Second inhabitant" since #496 landed first. Ready for re-review when you have a moment.

wshobson (changes requested):
Thanks for the follow-ups — the three blockers from my earlier review are all addressed (marketplace.json unicode, `$ARGUMENTS` handling, and the README clarification on unsigned approval records). Nicely done.

One new blocker that surfaced from #503: `hooks/hooks.json` here uses `"hook": {...}` (singular) inside each matcher, but current Claude Code expects `"hooks": [...]` (array). It's the same schema bug #503 fixes in protect-mcp, and the plugin will fail to load on current CC until it's converted. Easiest path: wait for #503 to land and rebase, or just preemptively switch both PreToolUse and PostToolUse blocks to the array form here.

Minor: the PR body says "9 rules; 4 forbid + 1 permit" but `policies/review-agent-governance.cedar` actually has 5 forbid + 1 permit = 6. The README's verbal coverage is correct, only the numeric description is off.

Happy to land this once the schema is fixed.

tomjwxf:
Thanks @wshobson, both items addressed in 02e77c5 (force-pushed).

**1. hooks.json schema fix.** Converted both `PreToolUse` and `PostToolUse`
blocks from the deprecated singular `"hook": {...}` shape to the current
`"hooks": [{...}]` array shape, matching the conversion #503 just landed
in `plugins/protect-mcp/hooks/hooks.json`. Confirmed the resulting
structure is byte-identical at the schema level to the post-#503
protect-mcp file (same key shape, same nesting depth, just the per-plugin
command strings differ). The `if [ -f ... ]` short-circuit and the
`@0.5.5` pin both survived the conversion intact.

**2. PR body rule-count correction.** You're right, the policy file has
5 forbid + 1 permit = 6 rules, not the "9 rules; 4 forbid + 1 permit"
the original PR description claimed. The README's verbal coverage list
matches the actual rules, so this is purely a description-vs-file mismatch
in the PR body. I'll correct the body inline once we're closer to merge
so I'm not editing it under your in-flight review.

Also rebased onto current main as part of this push, so #503, #496, and
the recent supply-chain hardening from #499 are all in the base.
`mergeStateStatus` is back to clean.

CI is re-running on the force-push; will leave the change set as-is for
your re-review.

---

**Follow-ups (no action needed on this PR; flagging in case any of them
are useful)**:

- **Welcoming category description.** Picking up the non-blocking note
  about the category remaining open to non-Cedar / non-Ed25519
  governance-adjacent plugins: rather than wait for a non-ScopeBlind
  submission to trigger the change, I'm happy to send a small follow-up
  PR that softens the category description in `marketplace.json` to read
  as "cryptographic or procedural governance for AI agents (policy
  enforcement, signed audit trails, human-gate patterns, and related
  accountability tooling)" rather than the current more stack-specific
  framing. That preempts the situation where someone with a different
  stack reads the category as ScopeBlind-coded and goes elsewhere.

- **v0.2 scoped approvals.** As mentioned in the original PR body, the
  current `./.review-approved` flag pattern is intentionally simple for
  v0.1. v0.2 could add scoped approvals (`/approve-review --tool gh
  --count 1` for a one-shot, or `--until 5m` for a time window) so the
  approval semantics get richer without changing the receipt format.
  Separate PR if you'd want it.

- **Pin `@veritasacta/verify` references in docs.** The README and
  SKILL.md examples currently mention `npx @veritasacta/verify` without
  a pin. Given how strict you've been (correctly) about runtime pinning
  in #497 and on this PR, would you prefer those documentation snippets
  pinned to `@veritasacta/verify@0.3.0` too, even though they're
  pattern-illustrative rather than executed by the plugin runtime? I
  initially treated them as documentation-pattern (where unversioned
  reads better as "this is the CLI you reach for") but if you'd prefer
  them pinned for consistency with the runtime story, that's a one-liner
  follow-up.

Happy to land any of the three above as separate PRs after this one
merges. None blocks the current change.

wshobson (approved):
Re-reviewed after the rebase + schema conversion — all prior blockers resolved. CI is green, hooks.json matches the current array schema, `protect-mcp@0.5.5` is pinned, marketplace.json appends cleanly.

One small doc drift I went ahead and fixed for you in 1ca21d8 (pushed to your branch via maintainer edits): the "Composing with protect-mcp" example block in `skills/review-agent-setup/SKILL.md` was still using the deprecated `"hook": {...}` singular schema and `npx protect-mcp@latest` (unpinned). A user copy-pasting that into their own hooks.json would have hit exactly the schema-load failure #503 fixed for the runtime side. Now matches the array form and the @0.5.5 pin.

Approving conditional on the eval/JSON checks passing on the new commit. Will merge once they're green.

The PR-body rule-count nit (5 forbid + 1 permit, not 9) is fine to clean up post-merge as you suggested.

plugins/review-agent-governance/policies/review-agent-governance.cedar
```diff
@@ -0,0 +1,103 @@
...
+        "gh pr close",
+        "gh pr merge",
+        "gh pr edit",
+        "gh issue comment",
+        "gh issue close",
+        "gh issue edit",
+        "gh release create",
+        "gh release edit",
+        "gh api repos"
+    ]
```
victornicolet, line 35: Those policies will always error and never be considered in a authorization decision, same as the policies below that use the pattern `x in ["string1", "string2", ...]`. 
The Cedar `in` operator is meant to evaluate membership in hierarchies of entities, see documentation: https://docs.cedarpolicy.com/policies/syntax-operators.html#operator-in.
This is what https://github.com/wshobson/agents/issues/598 is pointing out.

tomjwxf: Landed: protect-mcp 0.7.0 fails closed, evaluates Cedar correctly against the current engine, and ships a self-test that proves an `in`-on-String policy denies rather than permit-all. Thanks again for flagging it.

PR #496, opened 2026-04-17
feat: add signed-audit-trails teaching plugin (third governance-category skill)

tomjwxf:
Hi @wshobson, third plugin in the `governance` category, this one is a
teaching skill rather than a runtime hook. Pairs naturally with the existing
`protect-mcp` (runtime) and `review-agent-governance` (runtime + human
approval) plugins without overlapping them.

## What this is

A single skill file, ~1,600 words, walking through signed-audit-trail setup
end-to-end: Cedar policy authoring, hook configuration, first receipt,
offline verification with `@veritasacta/verify`, live tamper-detection
demonstration, and CI/CD integration. Plus: the three cryptographic
invariants (JCS + Ed25519 + hash chain), a cross-implementation interop
table, SLSA composition via ResourceDescriptor byproduct, and common
pitfalls.

## Why a teaching plugin alongside the two runtime ones

- `protect-mcp` is the production runtime. Install it, hooks fire, receipts
  write.
- `review-agent-governance` is a focused runtime for review-surface gating.
- `signed-audit-trails` (this one) is for a different user: someone who has
  not yet decided to commit to the infrastructure and wants to understand
  the pattern in-session before proceeding.

A user invokes the skill via `Skill` in Claude Code, gets the walkthrough,
reproduces the demo locally, then either installs `protect-mcp` for
production use or decides receipts are not right for them. Education before
commitment.

## Files in the PR

```
plugins/signed-audit-trails/
├── .claude-plugin/plugin.json
├── README.md
└── skills/signed-audit-trails-recipe/SKILL.md   (~1600 words)
```

Plus one entry added to `.claude-plugin/marketplace.json`, right after
`protect-mcp`, keeping the governance category ordered by install-then-learn
for protect-mcp and learn-then-install for signed-audit-trails.

## Verification

- [x] `python3 -m json.tool` validates `plugin.json` and `marketplace.json`
- [x] SKILL.md frontmatter matches the existing convention in this repo
      (`name`, `description`, YAML delimiters)
- [x] README.md mirrors the structure of `protect-mcp/README.md` and
      `review-agent-governance/README.md`
- [x] All external links resolve (Cedar docs, RFCs, IETF draft, npm packages)
- [x] Keyword list distinguishes teaching content from runtime
      (`tutorial`, `skill`, `recipe`)

## Notes

- This is additive. No changes to `protect-mcp` or `review-agent-governance`.
- The SKILL is self-contained. It references the runtime plugins but does
  not require them to be installed; a user can read the skill, run the
  demo with `npx`, and never touch the runtime plugin if they choose not
  to.
- Keywords include both governance (category match) and tutorial-class
  terms for discovery.

Thanks for the governance category in #484 and #495. This rounds out the
shelf with a learn-before-commit entry point for users evaluating the
pattern.

cc @wshobson

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -13,7 +13,7 @@
     {
       "name": "documentation-standards",
       "source": "./plugins/documentation-standards",
-      "description": "HADS (Human-AI Document Standard) — semantic tagging convention for writing docs that work efficiently for both humans and AI models. Reduces token consumption by separating machine-critical facts from human context.",
+      "description": "HADS (Human-AI Document Standard) \u2014 semantic tagging convention for writing docs that work efficiently for both humans and AI models. Reduces token consumption by separating machine-critical facts from human context.",
       "version": "1.0.0",
       "author": {
         "name": "Niksa Barlovic",
@@ -900,7 +900,7 @@
       "description": "Binary reverse engineering, malware analysis, firmware security, and software protection research for authorized security research, CTF competitions, and defensive security",
       "version": "1.0.0",
       "author": {
-        "name": "Dávid Balatoni",
+        "name": "D\u00e1vid Balatoni",
         "url": "https://github.com/balcsida"
       },
       "homepage": "https://github.com/wshobson/agents",
@@ -909,7 +909,7 @@
     },
     {
       "name": "conductor",
-      "description": "Context-Driven Development plugin that transforms Claude Code into a project management tool with structured workflow: Context → Spec & Plan → Implement",
+      "description": "Context-Driven Development plugin that transforms Claude Code into a project management tool with structured workflow: Context \u2192 Spec & Plan \u2192 Implement",
       "version": "1.2.1",
       "author": {
         "name": "Seth Hobson",
@@ -1002,7 +1002,7 @@
         "url": "https://github.com/Anasss/qa-orchestra.git",
         "path": "."
       },
-      "description": "Multi-agent QA toolkit with 10 specialized agents covering the full QA lifecycle — orchestrator, environment-manager, functional-reviewer, test-scenario-designer, browser-validator, automation-writer, manual-validator, bug-reporter, release-analyzer, and smart-test-selector. Stack-agnostic, output-chained, designed around live validation via Chrome MCP.",
+      "description": "Multi-agent QA toolkit with 10 specialized agents covering the full QA lifecycle \u2014 orchestrator, environment-manager, functional-reviewer, test-scenario-designer, browser-validator, automation-writer, manual-validator, bug-reporter, release-analyzer, and smart-test-selector. Stack-agnostic, output-chained, designed around live validation via Chrome MCP.",
       "version": "1.0.0",
       "author": {
         "name": "Anass Rach",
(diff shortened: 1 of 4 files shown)


wshobson (changes requested):
Thanks @tomjwxf — the SKILL reads well and the three-invariants explanation (JCS + Ed25519 + hash chain) plus the cross-implementation interop table are good teaching content. One blocker before merge:

**`marketplace.json` unicode regression (blocking).** Same issue as #495: the diff rewrites literal UTF-8 characters on other, unrelated plugin entries as ASCII escapes (`\u2014`, `\u2192`, `\u00e1`). Please only append the new `signed-audit-trails` entry and leave the surrounding descriptions byte-for-byte intact. Likely a `json.dump` round-trip without `ensure_ascii=False`.

**Also:** this PR and #495 both modify the same section of `marketplace.json` and will conflict. Once the unicode issue is fixed on both, one will need to rebase on the other.

**Non-blocking thoughts:**

- The `refs.arewm.com/agent-commit/v0.2` link should probably be verified before merge.
- Worth deciding with the maintainer whether a teaching plugin with no agents/commands/hooks — just one SKILL — is better as a standalone plugin or as an additional skill under `plugins/protect-mcp/skills/`. Both are defensible; the current decoupling has the advantage that a user evaluating receipts before committing to infrastructure can install this one alone. Your call to propose, maintainer's call to decide.
- The walkthrough shows `--fail-on-missing-policy false` and then warns against it for production — good. Showing the strict form end-to-end as a second example would tighten the "production-ready" pathway.

Happy to re-review once the `marketplace.json` diff is clean.

tomjwxf:
Thanks @wshobson, pushed 34b0c54 addressing the blocker plus the non-blocking notes:

**1. marketplace.json unicode regression — fixed.** Same fix as #495: reset the file to upstream HEAD and did a string-based append of only the new `signed-audit-trails` entry. No other entries modified. `grep -c "\\u" marketplace.json` = 0. Root cause was `json.dump` without `ensure_ascii=False`; the string-based approach sidesteps that entirely so it does not regress on a future edit either.

**2. refs.arewm.com/agent-commit/v0.2 — verified live.** `curl -sIL` returns HTTP 200. Arewm pushed v0.2 a few hours ago, and my companion PR at [arewm/refs.arewm.com#2](https://github.com/arewm/refs.arewm.com/pull/2) builds on v0.2 specifically. Link is correct.

**3. On the standalone-plugin vs nested-skill question — your call, but here is my read.** The user I had in mind for this plugin is someone evaluating the receipt pattern BEFORE they install `protect-mcp`. If the teaching skill lives under `plugins/protect-mcp/skills/`, the user has to install the runtime to read the skill, which is the opposite of the "learn before commit" path. Keeping it as a standalone plugin lets the user invoke the skill via `Skill`, reproduce the demo with `npx` (no runtime install), and only THEN install `protect-mcp` if they decide to proceed. Happy to restructure if you prefer the nested shape, but that is the UX argument for standalone.

**4. Strict-form `--fail-on-missing-policy true` example — deferred.** Agreed this would tighten the production-ready pathway. Not in this fix because it is new content, not a bug fix. Will add it as a follow-up SKILL revision once this merges, alongside the pinned-`protect-mcp@x.y.z` change I mentioned on #495.

**On the #495 conflict:** whichever you want to land first is fine. The string-based insertion approach in both PRs means the rebase of whichever-is-second is mechanical (just re-insert after the new first entry).

Re-requesting review.

PR #503, opened 2026-04-23
fix(protect-mcp): use current hook schema (hooks array)  

DiyRex:
The protect-mcp hooks.json uses an outdated schema with `"hook": {...}` (singular object).                                                                                        Current Claude Code expects `"hooks": [...]` (array of hook objects) inside each matcher.                                                                                         
                                                                                                                                                                                    
 On load, Claude Code reports:                                                                                                                                                     
    protect-mcp@claude-code-workflows [protect-mcp]: Hook load failed:                                                                                                              
    path: hooks.PreToolUse.0.hooks — expected array, received undefined                                                                                                             
    path: hooks.PostToolUse.0.hooks — expected array, received undefined   

This PR renames `hook` → `hooks` and wraps each command object in an array,                                                                                                       
 matching the schema used by other plugins in this marketplace (e.g. block-no-verify).

### DIFF
--- a/plugins/protect-mcp/hooks/hooks.json
+++ b/plugins/protect-mcp/hooks/hooks.json
@@ -3,19 +3,23 @@
     "PreToolUse": [
       {
         "matcher": ".*",
-        "hook": {
-          "type": "command",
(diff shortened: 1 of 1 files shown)


wshobson:
Verified against the current hooks schema in the official docs — the array form is correct, and this fixes a real load failure. Merging. Thanks!

PR #530, opened 2026-05-13
fix(plugin-eval): broaden MISSING_TRIGGER pattern to match canonical phrasings

jondwillis:
## Problem

The Layer-1 `MISSING_TRIGGER` heuristic in `plugins/plugin-eval/src/plugin_eval/layers/static.py:132` is a literal substring match against `Use when` / `Use this skill when` / `Use proactively` / `Trigger when`. This produces large numbers of false-positive flags against real, well-formed plugins because three common phrasings don't contain the literal `Use when`:

1. **Third-person canonical form** — Anthropic's own `plugin-dev` plugin (specifically the `skill-development` skill) recommends `This skill should be used when …` as the *correct* description style and explicitly calls out the imperative `Use this skill when …` form as a mistake. The canonical phrasing contains `used when`, not `Use when`.

2. **Prepositional temporal triggers** — `Use after …`, `Use before …`, `Use immediately before …`, `Use whenever …`. These are common in self-audit, hook-adjacent, and lifecycle skills where the trigger is a temporal context rather than a generic "when".

3. **Path-triggered self-documentation** — `Auto-loads when …` for skills that use the `paths:` frontmatter for auto-loading and describe that mechanism in their description.

## Evidence

Running plugin-eval against Anthropic's own plugin-dev (the authoritative source on how to write Claude Code plugin skills) produces **7 false-positive `MISSING_TRIGGER` flags** out of 7 skills — every one. The static linter penalises the very phrasing pattern plugin-dev's own skill-development guide recommends.

```text
# Before this PR
plugin-dev composite: 33.82  penalty: 0.50  (7× MISSING_TRIGGER)

# After this PR
plugin-dev composite: 55.18  penalty: 0.75  (0× MISSING_TRIGGER; 5 unrelated flags remain)
```

A separate test against the `functional-emotions` plugin shows a similar pattern: 6 of 11 skills flagged before, 0 after.

## Fix

- Extract the trigger regex to a module-level `_TRIGGER_PATTERN` constant so the anti-pattern detector and `_description_pushiness` heuristic can share it (previously two slightly-divergent copies).
- Expand the pattern to match `(?:should\s+be\s+)?use(?:d)?\s+(?:this\s+skill\s+)?(?:immediately\s+)?(?:when|after|before|whenever)` plus `auto-loads when` and `trigger when`.
- Keep all existing accepted phrasings (`Use when`, `Use this skill when`, `Use proactively`, `Trigger when`).

## Test plan

- [x] 22 tests pass (`uv run --extra dev pytest tests/test_static.py -v`), including:
  - 11 new parametrised positive cases covering the expanded forms
  - 4 parametrised negative cases (descriptions with no trigger phrase still flag)
  - 2 regression tests asserting that the canonical Anthropic phrasings (`This skill should be used when …`, `Use immediately before …`) do **not** produce `MISSING_TRIGGER`
- [x] End-to-end run against plugin-dev confirms 7 false positives → 0
- [x] End-to-end run against an external plugin confirms 6 false positives → 0

### DIFF
--- a/plugins/plugin-eval/src/plugin_eval/layers/static.py
+++ b/plugins/plugin-eval/src/plugin_eval/layers/static.py
@@ -30,6 +30,25 @@ def anti_pattern_penalty(count: int) -> float:
 # Line count threshold for BLOATED_SKILL (no references/ dir)
 _BLOATED_LINE_THRESHOLD = 800
 
+# Canonical trigger phrasings the model can use to decide when to invoke a skill.
+# Matches:
+#   - Imperative second-person: "Use when …", "Use this skill when …"
+#   - Third-person canonical (Anthropic plugin-dev recommends this form):
+#     "This skill should be used when …", "Used when …"
+#   - Prepositional temporal triggers commonly used in self-audit / hook-adjacent
+#     skills: "Use after …", "Use before …", "Use immediately before …",
+#     "Use whenever …", "Used after …", etc.
+#   - Explicit auto-load self-documentation: "Auto-loads when …"
+#   - Existing explicit markers: "Use proactively", "Trigger when …"
+_TRIGGER_PATTERN = re.compile(
+    r"\b(?:should\s+be\s+)?used?\s+(?:this\s+skill\s+)?(?:immediately\s+)?"
+    r"(?:when|after|before|whenever)\b"
+    r"|\buse\s+proactively\b"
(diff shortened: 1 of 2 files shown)


wshobson:
@jondwillis Thanks again! On a roll, my man. 👊

PR #531, opened 2026-05-13
fix(plugin-eval): exempt slash-only and path-triggered skills from MISSING_TRIGGER

jondwillis:
## Problem

The `MISSING_TRIGGER` anti-pattern check assumes every skill is auto-invoked by the model based on its description. Claude Code actually supports three invocation mechanisms:

1. **Description-driven auto-invocation** — the default. `MISSING_TRIGGER` is the correct gate.
2. **Slash-only invocation** — `disable-model-invocation: true` in the SKILL.md frontmatter. The model cannot auto-invoke these, so a trigger phrase in the description is irrelevant.
3. **Path-triggered auto-load** — `paths:` frontmatter glob. The skill loads when the model opens a matching file. The description is documentation rather than a discovery surface; description-level trigger phrasing is orthogonal.

The current check fires against every skill, producing false-positive flags against operator-only commands (setup, doctor, audit tools) and against path-triggered guards (test-edit primers, etc.).

## Fix

- Add a `_skill_uses_description_trigger(skill)` predicate that inspects `skill.frontmatter` for `disable-model-invocation: true` and a non-empty `paths:` value.
- Run the existing `MISSING_TRIGGER` check only when the predicate returns `True`.
- Behaviour for ordinary model-invocable skills is unchanged.

Note: the parser already reads the full frontmatter dict into `ParsedSkill.frontmatter`, so this change is purely additive — no parsing changes required.

## Test plan

- [x] 4 new tests in `tests/test_static.py::TestTriggerExemptions`:
  - `test_disable_model_invocation_exempts_skill` — slash-only skill without a trigger phrase is not flagged
  - `test_paths_auto_load_exempts_skill` — `paths:` skill without a trigger phrase is not flagged
  - `test_disable_model_invocation_false_still_checks_trigger` — explicit `false` does not exempt
  - `test_empty_paths_value_still_checks_trigger` — empty-string `paths:` is not a valid auto-load configuration
- [x] Full plugin-eval suite (73 tests) passes via `uv run --extra dev pytest tests/ -v`.

## Relation to other PRs

Conceptually independent from #530 (broaden MISSING_TRIGGER regex). The two can land in either order; together they eliminate the remaining classes of false-positive MISSING_TRIGGER flags against real-world plugins.

### DIFF
--- a/plugins/plugin-eval/src/plugin_eval/layers/static.py
+++ b/plugins/plugin-eval/src/plugin_eval/layers/static.py
@@ -32,6 +32,23 @@ def anti_pattern_penalty(count: int) -> float:
     return max(0.5, 1.0 - 0.05 * count)
 
 
+def _skill_uses_description_trigger(skill: ParsedSkill) -> bool:
+    """Return True if the skill relies on its description to be auto-invoked.
+
+    Skills that opt out of model-driven invocation (`disable-model-invocation:
+    true`) or that auto-load on a path glob (`paths:` frontmatter) use a
+    different trigger mechanism and should not be checked for a "Use when …"
+    phrase in the description.
+    """
+    fm = skill.frontmatter
+    if fm.get("disable-model-invocation") is True:
+        return False
+    paths = fm.get("paths")
+    if paths is not None and paths != "":
+        return False
+    return True
+
+
 # Line count threshold for BLOATED_SKILL (no references/ dir)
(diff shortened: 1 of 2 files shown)


wshobson:
@coderabbitai full review

wshobson:
@jondwillis Got you covered. Thanks again for this fix.

PR #532, opened 2026-05-13
fix(plugin-eval): surface plugin-level depth downgrades loudly

jondwillis:
## Problem

`plugin-eval certify <plugin-dir>` is documented and named as a deep, three-layer evaluation (static + judge + Monte Carlo). In practice, plugin-level evaluation only runs the static layer regardless of the requested depth — the judge and Monte Carlo analyzers are per-skill primitives, and `EvalEngine.evaluate_plugin` hard-codes a static-only path:

```python
# engine.py
def evaluate_plugin(self, plugin_dir: Path) -> PluginEvalResult:
    """...
    Note: Plugin-level evaluation currently only runs Layer 1 (static).
    Judge and Monte Carlo require per-skill evaluation. The confidence
    label is always "Estimated" regardless of requested depth.
    """
    # ...only Layer 1 (static) is run...
    composite.confidence_label = Depth.QUICK.confidence_label
```

Nothing surfaces this to the user. Running `plugin-eval certify <plugin-dir> --output markdown` produces a report that reads `Depth: deep` and exits in seconds, with the downgrade signal buried in:
- a `Confidence: Estimated` row in the score table, and
- a `No model usage (static-only evaluation)` footnote near the end.

A consumer comparing scores across plugins will quietly mistake an in-effect static-only run for a deep certification.

## Fix

This PR makes the downgrade impossible to miss without changing the underlying evaluation pipeline (per-skill judge aggregation is a larger feature, not a bug fix).

1. **CLI warning to stderr.** `_run_score` detects a plugin-target run at non-quick depth and prints a yellow `warning:` line to stderr naming the skipped layers and the workaround (run on a single skill to get the deeper layers).

2. **In-band markdown callout.** The reporter now derives the *effective* depth from the set of layers actually present in the result. When effective differs from requested, the report header reads
   ```
   **Depth:** deep (requested) → quick (effective)
   ```
   and a `> Note:` block above the score table explains why and how to get the deeper layers.

3. **Effective-depth helper.** `_effective_depth(result)` maps the set of layer names (`static` / `judge` / `monte_carlo`) back to a `Depth` value, so the reporter never has to trust `result.config.depth` when describing what actually ran.

## Example output

Before:
```text
**Depth:** deep
[score table]
_No model usage (static-only evaluation)._
```

After (stderr):
```text
warning: plugin-level evaluation only runs the static layer; judge and Monte
Carlo layers require per-skill evaluation. Requested depth deep will be served
from the static layer only — confidence label will be Estimated regardless. To
use the deeper layers, point at an individual skill directory.
```

After (stdout markdown):
```text
**Depth:** deep (requested) → quick (effective)

> **Note:** Requested depth `deep` was downgraded to `quick` because
> plugin-level evaluation only runs the static layer. Judge and Monte Carlo
> layers require per-skill evaluation — point at an individual skill directory
> to use the deeper layers. Composite score and confidence reflect the layers
> actually run.
```

## Test plan

- [x] `TestDepthDowngradeWarning` (4 tests in `tests/test_reporter.py`): asserts the effective-depth helper, the "no warning when depth was honored" path, and the warning content for both deep and standard plugin-level requests.
- [x] `TestCLI` (2 new tests in `tests/test_cli.py`): asserts the stderr warning is emitted on plugin-level certify and is *not* emitted at quick depth.
- [x] Full plugin-eval suite (75 tests) passes via `uv run --extra dev pytest tests/`.

## Follow-up (out of scope for this PR)

A larger change could implement per-skill judge / Monte Carlo aggregation so plugin-level deep evaluation is honoured directly. That has architectural questions (aggregation strategy, cost handling, dimension reduction) that this PR deliberately doesn't try to settle. The warning + in-band note here is the minimal fix for the silent-downgrade bug; deeper-layer plugin aggregation can be a follow-up.

### DIFF
--- a/plugins/plugin-eval/src/plugin_eval/cli.py
+++ b/plugins/plugin-eval/src/plugin_eval/cli.py
@@ -17,6 +17,7 @@ app = typer.Typer(
     add_completion=False,
 )
 console = Console()
+stderr_console = Console(stderr=True)
 
 
 def _detect_target(path: Path) -> str:
@@ -55,6 +56,15 @@ def _run_score(
     if target == "skill":
         result = engine.evaluate_skill(path)
     elif target == "plugin":
+        if depth != Depth.QUICK:
+            stderr_console.print(
+                f"[yellow]warning:[/yellow] plugin-level evaluation only runs the "
+                f"static layer; judge and Monte Carlo layers require per-skill "
+                f"evaluation. Requested depth [bold]{depth.value}[/bold] will be "
+                f"served from the static layer only — confidence label will be "
+                f"[bold]Estimated[/bold] regardless. To use the deeper layers, "
+                f"point at an individual skill directory."
+            )
         result = engine.evaluate_plugin(path)
     else:
         # Attempt skill evaluation as fallback
--- a/plugins/plugin-eval/src/plugin_eval/reporter.py
+++ b/plugins/plugin-eval/src/plugin_eval/reporter.py
@@ -2,7 +2,20 @@
 
 from __future__ import annotations
 
-from plugin_eval.models import PluginEvalResult
+from plugin_eval.models import Depth, PluginEvalResult
+
+
+_LAYER_TO_DEPTH: dict[frozenset[str], Depth] = {
+    frozenset({"static", "judge", "monte_carlo"}): Depth.DEEP,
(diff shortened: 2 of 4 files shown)


wshobson:
@jondwillis Thanks! Look great.

PR #576, opened 2026-06-11
Add file-conversion plugin (ChangeThisFile)

aadilr:
Adds a `file-conversion` plugin under **utilities** — converts files between 999 routes (PDF↔Word, HEIC→JPG, MP4→MP3, CSV→JSON, EPUB→MOBI, and more across images, video, audio, documents, data, fonts, ebooks, archives) via the free [ChangeThisFile](https://changethisfile.com) API. No API key or signup.

- One skill: `file-conversion` — MCP-aware (uses `changethisfile:convert_file` when connected) with a zero-dependency `curl` script fallback (`scripts/convert.sh`), so it works in every harness this marketplace targets.
- Upstream repo: https://github.com/aadilr/changethisfile-mcp (MIT; also in the official MCP registry as `com.changethisfile/mcp`).
- Script tested end-to-end (md→pdf, csv→xlsx, clean error surfaces for unsupported routes; 25MB/5-per-min limits documented in the skill).
- `make generate-all` run (adapters regenerated), `make validate STRICT=1` passes across all 5 harnesses; `make garden` warnings are all pre-existing in other plugins (SKILL_OVER_CODEX_CAP), none introduced — our SKILL.md body is well under the Codex cap.

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -482,6 +482,18 @@
       "category": "Coding"
     },
     {
+      "name": "file-conversion",
+      "source": {
+        "source": "local",
+        "path": "./plugins/file-conversion"
+      },
+      "policy": {
+        "installation": "AVAILABLE",
+        "authentication": "ON_USE"
+      },
+      "category": "Coding"
+    },
+    {
       "name": "framework-migration",
       "source": {
         "source": "local",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -13,7 +13,7 @@
     {
       "name": "documentation-standards",
       "source": "./plugins/documentation-standards",
-      "description": "HADS (Human-AI Document Standard) — semantic tagging convention for writing docs that work efficiently for both humans and AI models. Reduces token consumption by separating machine-critical facts from human context.",
+      "description": "HADS (Human-AI Document Standard) \u2014 semantic tagging convention for writing docs that work efficiently for both humans and AI models. Reduces token consumption by separating machine-critical facts from human context.",
       "version": "1.0.1",
       "author": {
         "name": "Niksa Barlovic",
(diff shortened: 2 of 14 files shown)


aadilr:
Thanks for the review — addressed all four findings in 6cc85c9:

- **`.claude-plugin/marketplace.json`**: reverted the incidental unicode-escape normalizations on unrelated entries; only the `file-conversion` entry remains (kept the registry's existing `\uXXXX` escaping).
- **`.cursor-plugin/**`**: removed the hand edits and regenerated via `make generate-all` (`make validate` → OK across all 5 harnesses).
- **`convert.sh` path traversal**: target format is now validated (alphanumeric only) and the output path rejects `..` before `curl -o`.
- **`convert.sh` JSON escaping**: the JSON-RPC payload is now built with `jq --arg` so arbitrary filenames are escaped safely, with a strict printable-ASCII fallback when `jq` is absent.

wshobson:
Thanks for a well-put-together submission — the regenerated registries are spot-on and CI is fully green. Two things before merge:

1. `scripts/convert.sh` passes the whole base64 payload to `jq --arg`, which hits the OS argv limit (~96 KB per arg on Linux) — feeding it via `--rawfile`/stdin will make larger files work as advertised.
2. Please add the plugin to `docs/plugins.md` (Utilities) and bump the counts in README.md/AGENTS.md.

Happy to re-review after that.

aadilr:
Both addressed:

1. **base64 argv limit** — `convert.sh` now writes the payload to a temp file and reads it into jq via `--rawfile` instead of `--arg`, so inputs past ~96KB work as advertised (verified with a 164KB file → 219KB base64, which failed before).
2. **docs** — added `file-conversion` to `docs/plugins.md` (Utilities) and bumped the count 84→85 (83 local + 2 external) across README, AGENTS, ARCHITECTURE, and the docs pages.

Also hardened the output-path check to reject absolute/`~` paths (per CodeRabbit), with the auto-derived default still handling absolute inputs. markdownlint green locally. Thanks for the thorough review!

Issue #598, opened 2026-06-24
Security: `in` operator on String context attribute silently discards all forbid rules in review-agent-governance.cedar


matiaszabal:
## Security Notice — Cedar `in`-on-String silent policy discard

This is a responsible disclosure of a Cedar authorization hazard affecting `plugins/review-agent-governance/policies/review-agent-governance.cedar` (and its copies in at least 4 other repositories).

**Summary:** All 5 forbid rules in this policy use `context.x in [string_list]`, which is Cedar's entity-hierarchy operator applied to a String context attribute. Cedar's Rust engine raises a runtime type error and — per Cedar's documented error-ignorance semantics — silently discards the erroring forbid rule. The terminal `permit (principal, action, resource)` then fires for every request. **Every governance gate in this policy is a no-op.**

**Upstream disclosure:** This hazard has been filed with the Cedar project: cedar-policy/cedar#2428

**Confirmation (cedar-python 0.1.4 / cedar-policy 4.8.2):**

| Test | Intent | Actual |
|------|--------|--------|
| `command_pattern = "gh pr merge"` | DENY | **PERMIT** |
| `command_pattern = "git push --force"`, `target_branch = "main"` | DENY | **PERMIT** |
| `path_starts_with = ".github/workflows/"` (Write action) | DENY | **PERMIT** |
| `method = "POST"`, `url_host = "api.github.com"` (WebFetch) | DENY | **PERMIT** |

5/5 governance gates bypassed (100%).

**Type error string:** `error while evaluating policy 'policy0': type error: expected (entity of type 'any_entity_type'), got string`

**Fix:** Replace `context.x in [list]` with `[list].contains(context.x)` in every forbid rule.

```cedar
// Current (broken — in-on-String, silently discarded):
forbid (...) when { context.command_pattern in ["gh pr merge", ...] };

// Fixed:
forbid (...) when { ["gh pr merge", ...].contains(context.command_pattern) };
```

This policy is distributed as a reusable Claude Code plugin and has been copied into at least 4 additional public repositories (`bachsh/supermarket`, `Jhabbig/Habbig`, `AdmiralGallade/harness-repository`, `steph-frtech/claude-code-up`). Users of those copies are also affected.

The upstream Cedar issue (cedar-policy/cedar#2428) proposes schemaless load-time warnings and a lint mode to prevent this class of mistake in the future.

wshobson:
[...]

Issue #591, opened 2026-06-21
[BUG] use plugin-eval standard but cannot output with llm


lkjie:
### Preliminary Checks

- [x] I have read the [Code of Conduct](https://github.com/wshobson/agents/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] I have searched existing issues to ensure this is not a duplicate
- [x] This report contains only technical information about a bug

### Harness

Claude Code

### Harness version

2.1.185 (Claude Code)

### Affected plugin / agent / skill / command

plugins/plugin-eval

### Bug description

# lkjie @ openubmc-dev in ~/projects/rd_lab_skill_eval/repos/agents/plugins/plugin-eval on git:main x .venv [20:36:55] 
$ uv run plugin-eval score /home/lkjie/projects/rd_lab_skill_eval/repos/agents/plugins/protect-mcp/skills/protect-mcp-setup --depth standard --output json > /home/lkjie/projects/rd_lab_skill_eval/work/plugin_eval/result.json

json result
```json
{
  "plugin_path": "/home/lkjie/projects/rd_lab_skill_eval/repos/agents/plugins/protect-mcp/skills/protect-mcp-setup",
  "timestamp": "2026-06-21T11:50:00.126221+00:00",
  "config": {
    "depth": "standard",
    "concurrency": 4,
    "model_tier": "auto",
    "output_format": "json",
    "verbose": false,
    "corpus_path": null,
    "auth": "max",
    "judges": 1,
    "monte_carlo_n": null
  },
  "layers": [
    {
      "layer": "static",
      "score": 0.7561739130434783,
      "sub_scores": {
        "frontmatter_quality": 0.6499999999999999,
        "orchestration_wiring": 0.9,
        "progressive_disclosure": 0.6,
        "structural_completeness": 0.7000000000000001,
        "token_efficiency": 0.9130434782608696,
        "ecosystem_coherence": 0.75,
        "harness_portability": 1.0
      },
      "anti_patterns": [],
      "metadata": {
        "skill_name": "protect-mcp-setup",
        "line_count": 208
      }
    },
    {
      "layer": "judge",
      "score": 0.5,
      "sub_scores": {
        "triggering_accuracy": 0.5,
        "orchestration_fitness": 0.5,
        "output_quality": 0.5,
        "scope_calibration": 0.5
      },
      "anti_patterns": [],
      "metadata": {
        "triggering": {
          "raw": "",
          "score": 0.5
        },
        "orchestration": {
[...]

PR #600, opened 2026-06-25
fix: plugin-eval judge silent-failure (#591) + review-agent-governance Cedar hardening (#598)

wshobson:
## Summary

Fixes two open issues, plus surfaces a third (filed separately).

### #591 — plugin-eval LLM judge silently emitted fabricated 0.5 scores
`query_llm` (and the Monte Carlo `run_simulation`) read response text from `ResultMessage.content`, which does not exist on that `claude-agent-sdk` type — so the judge always saw empty text, JSON parsing always failed, and every dimension silently fell back to `0.5` (grade **F**) while the command exited 0. The judge carries ~72% of the rubric weight, so genuinely strong skills were reported as failing.

- Read assistant text the documented way — `AssistantMessage` → `TextBlock.text`, with `ResultMessage.result` as fallback — in both the judge and Monte Carlo layers (shared `layers/_sdk.py::collect_sdk_output`).
- Genuine measurement failures (errored / empty / unparseable, or SDK missing) now surface as **"unmeasured"** — the dimension is omitted from `sub_scores` so the composite engine recomputes static-only, and the CLI prints a clear stderr warning — instead of fabricating `0.5`.
- Removed dead `auth` / `model_tier` config that was threaded through but never used.
- Added real (non-mocked) extraction tests that exercise the actual SDK-message path that let this bug ship.

### #598 — review-agent-governance Cedar `in`-on-String silently disabled forbid rules
All 5 `forbid` rules used `context.<stringAttr> in [strings]`. In Cedar `in` requires an **entity** LHS, so this is a type error the engine discards — leaving the terminal `permit` to fire. The reporter is essentially correct (the runtime evaluates Cedar via the real engine).

- Rewrote all 5 rules to the validator-clean set form `["s1","s2"].contains(context.<attr>)` (legitimate entity-`in` on the `action` scope left untouched).
- Added a `.cedarschema` typing the context attributes as `String` so `cedar validate` catches this class of mistake at load time.
- Added a regression test (`test/run-tests.sh`): Part A (always runs) greps for the bug pattern; Part B runs `cedar validate` when the CLI is present.

### ⚠️ Separate finding surfaced while verifying #598
The plugin's `hooks/hooks.json` invokes `protect-mcp@0.5.5 evaluate` / `sign` subcommands that **do not exist** in that version (its CLI is `serve` / `init-hooks` / `simulate` / `--cedar`). So the governance gate does not run as shipped, independent of the Cedar syntax. This is a larger defect than #598 as filed and is being tracked as its own issue.

## Verification
- `make test` (plugin-eval + tools), `make validate STRICT=1`, `make garden`, and `make generate-all` drift check all pass; **122 plugin-eval tests pass**; `src/plugin_eval/` is ruff + format + ty clean.
- Cedar policy test passes (Part A); Part B skips gracefully where the `cedar` CLI is absent.

## Notes
- Relates to #591 and #598 (intentionally not auto-closing — the issue replies/closes and the new broken-hook issue are being handled separately).
- Design + plan: `docs/superpowers/specs/2026-06-25-open-issue-resolution-design.md`, `docs/superpowers/plans/2026-06-25-open-issue-resolution.md`.
- Known follow-ups (out of scope here): route the Monte Carlo layer through the same "unmeasured" omission as the judge; wire `cedar validate` into `make validate`; fix/regenerate the protect-mcp hook integration.

### DIFF
--- a/docs/plugin-eval.md
+++ b/docs/plugin-eval.md
@@ -91,7 +91,6 @@ uv run plugin-eval score path/to/skill --threshold 70
 | `--output` | `markdown` | `json`, `markdown`, `html` |
 | `--verbose` | `false` | Show detailed output |
 | `--concurrency` | `4` | Max concurrent LLM calls (1–20) |
-| `--auth` | `max` | Auth mode: `max` (Claude Code Max plan) or `api-key` |
 | `--threshold` | none | Minimum score; exit 1 if below |
 
 ### `certify` — Full certification with badge
--- a/docs/superpowers/plans/2026-06-25-open-issue-resolution.md
+++ b/docs/superpowers/plans/2026-06-25-open-issue-resolution.md
@@ -698,9 +698,11 @@ action "WebFetch" appliesTo {
 
 ```bash
 cd plugins/review-agent-governance/policies
-command -v cedar >/dev/null 2>&1 && \
-  cedar validate --policies review-agent-governance.cedar --schema review-agent-governance.cedarschema || \
+if command -v cedar >/dev/null 2>&1; then
+  cedar validate --policies review-agent-governance.cedar --schema review-agent-governance.cedarschema
+else
   echo "cedar CLI not present — schema is documentation/validation-only"
+fi
 ```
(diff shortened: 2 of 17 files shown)


wshobson:
@coderabbitai review

PR #602, opened 2026-06-28
Pin protect-mcp@0.7.4 in the governance plugins (closes #601)

### LINKED ISSUE #601: review-agent-governance: hooks invoke non-existent protect-mcp evaluate/sign subcommands (gate doesn't run on @0.5.5)
## Summary

The `review-agent-governance` plugin's hook integration targets `protect-mcp` subcommands that **do not exist** in the pinned version, so the governance gate does not run as shipped.

`plugins/review-agent-governance/hooks/hooks.json` invokes:
- PreToolUse: `npx protect-mcp@0.5.5 evaluate --policy … --tool … --input …`
- PostToolUse: `npx protect-mcp@0.5.5 sign …`

But `protect-mcp@0.5.5` has **no `evaluate` or `sign` subcommand**. Its actual CLI is:

```
protect-mcp serve [--port] [--enforce] [--policy] [--cedar <dir>]
protect-mcp init-hooks [--dir] [--port]
protect-mcp simulate --policy <path> [--log] [--tier] [--json]
protect-mcp quickstart | connect | init | demo | trace | status | digest | receipts | bundle | report
```

Running the shipped command falls through to protect-mcp's "wrapper" mode and errors (`Missing "--" separator before the command to wrap`). So the PreToolUse gate never enforces the policy, and PostToolUse never signs receipts.

## Impact

The plugin's core promise — gating consequential agent actions behind human approval, with Ed25519 receipts — does not function on `protect-mcp@0.5.5`. (This is separate from #598, which fixed the Cedar policy itself; even a correct policy isn't evaluated because the hook command is wrong.)

`protect-mcp`'s own `plugins/protect-mcp/test/run-tests.sh` is also affected — it invokes `protect-mcp@0.5.5 evaluate …` / `sign …`, so it would fail against the pinned version too.

## How it was found

Surfaced while verifying #598 (PR #600): a live `npx protect-mcp@0.5.5 evaluate …` run returned the wrapper-mode usage error, and `--help` confirmed the current subcommand set.

## Suggested fix

The modern protect-mcp integration is server-based: `protect-mcp serve` (HTTP hook server, default port 9377) with `--cedar <dir>` for local Cedar-via-WASM evaluation, and `protect-mcp init-hooks` generates the correct Claude Code hook config + sample policy.

- Regenerate `hooks/hooks.json` via `protect-mcp init-hooks` (or hand-port to the `serve` model), and update `plugins/protect-mcp/test/run-tests.sh` accordingly.
- Re-pin `protect-mcp` and verify the deny/permit flow end-to-end against the published CLI.

## Affected files
- `plugins/review-agent-governance/hooks/hooks.json`
- `plugins/protect-mcp/test/run-tests.sh` (same `evaluate`/`sign` assumption)

tomjwxf:
Both governance plugins invoke `protect-mcp evaluate` (PreToolUse) and `protect-mcp sign` (PostToolUse), but pinned `protect-mcp@0.5.5`, which predates those verbs. That is issue #601: the hooks call subcommands that did not exist in the pinned version. This bumps the pin in both `plugins/review-agent-governance/hooks/hooks.json` and `plugins/protect-mcp/hooks/hooks.json`, so the verbs resolve and the hooks run as written.

The pin is `0.7.4` (originally proposed as 0.7.0): `evaluate` passes the tool input through to Cedar as `context.input` from 0.7.4, which any command-scoping policy (including this repo's test policy) needs, and 0.7.4 also carries the per-host `--format` hook adapters. The 0.7.x line is a security release series: the Cedar gate fails closed on any evaluation error, evaluates Cedar correctly against cedar-wasm 4.x, and refuses to arm unless a startup self-test proves a known-forbidden vector is denied.

Review follow-ups (second commit):

- `plugins/protect-mcp/test/run-tests.sh`: pins updated in all 6 places; `keygen` (removed in 0.7.0) replaced with `init --dir`; receipts are read from `receipts.jsonl` (0.7.0+ appends one JSON per line); the offline verify step pins the current `@veritasacta/verify` and passes the signer public key explicitly (v2 receipts do not embed it).
- `fixtures/test-policy.cedar`: rewritten in the entity shape protect-mcp actually evaluates (`action == Action::"MCP::Tool::call"`, the tool as `resource`, input at `context.input`). The old `Action::"Read"` shape only appeared to work under 0.5.5's broken evaluator; under real fail-closed evaluation nothing matched and every allow test denied.
- `expected/receipt-schema.json`: updated to the v2 receipt envelope.
- `SKILL.md` copy-paste examples bumped in both plugins. The remaining `0.5.5` strings under `docs/superpowers/` are dated investigation records where 0.5.5 is historically accurate, so they were left as written.

Test suite: 8/8 locally via `plugins/protect-mcp/test/run-tests.sh`.

Note on the PostToolUse `sign` step: it is best-effort and never blocks a tool. If no signing key is configured it records an honest unsigned receipt line rather than failing. To produce signed receipts, run `protect-mcp init` once and point `--key` at `keys/gateway.json`.

Disclosure: I maintain protect-mcp upstream and originally authored these plugins, per the disclosure policy added in #611.

### DIFF
--- a/plugins/protect-mcp/hooks/hooks.json
+++ b/plugins/protect-mcp/hooks/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "npx protect-mcp@0.7.0 evaluate --policy \"${PROTECT_MCP_POLICY:-./protect.cedar}\" --tool \"$TOOL_NAME\" --input \"$TOOL_INPUT\" --fail-on-missing-policy false"
+            "command": "npx protect-mcp@0.7.4 evaluate --policy \"${PROTECT_MCP_POLICY:-./protect.cedar}\" --tool \"$TOOL_NAME\" --input \"$TOOL_INPUT\" --fail-on-missing-policy false"
           }
         ]
       }
@@ -17,7 +17,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "npx protect-mcp@0.7.0 sign --tool \"$TOOL_NAME\" --input \"$TOOL_INPUT\" --output \"$TOOL_OUTPUT\" --receipts \"${PROTECT_MCP_RECEIPTS:-./receipts/}\" --key \"${PROTECT_MCP_KEY:-./protect-mcp.key}\""
+            "command": "npx protect-mcp@0.7.4 sign --tool \"$TOOL_NAME\" --input \"$TOOL_INPUT\" --output \"$TOOL_OUTPUT\" --receipts \"${PROTECT_MCP_RECEIPTS:-./receipts/}\" --key \"${PROTECT_MCP_KEY:-./protect-mcp.key}\""
           }
         ]
       }
--- a/plugins/protect-mcp/hooks/hooks.json
+++ b/plugins/protect-mcp/hooks/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "npx protect-mcp@0.5.5 evaluate --policy \"${PROTECT_MCP_POLICY:-./protect.cedar}\" --tool \"$TOOL_NAME\" --input \"$TOOL_INPUT\" --fail-on-missing-policy false"
+            "command": "npx protect-mcp@0.7.0 evaluate --policy \"${PROTECT_MCP_POLICY:-./protect.cedar}\" --tool \"$TOOL_NAME\" --input \"$TOOL_INPUT\" --fail-on-missing-policy false"
           }
         ]
       }
@@ -17,7 +17,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "npx protect-mcp@0.5.5 sign --tool \"$TOOL_NAME\" --input \"$TOOL_INPUT\" --output \"$TOOL_OUTPUT\" --receipts \"${PROTECT_MCP_RECEIPTS:-./receipts/}\" --key \"${PROTECT_MCP_KEY:-./protect-mcp.key}\""
+            "command": "npx protect-mcp@0.7.0 sign --tool \"$TOOL_NAME\" --input \"$TOOL_INPUT\" --output \"$TOOL_OUTPUT\" --receipts \"${PROTECT_MCP_RECEIPTS:-./receipts/}\" --key \"${PROTECT_MCP_KEY:-./protect-mcp.key}\""
           }
         ]
       }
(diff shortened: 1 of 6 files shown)


wshobson:
Thanks for the quick fix and the detailed verification notes — the 0.7.0 claims checked out on our side too (verified `evaluate`/`sign` in the shipped CLI and the fail-closed behavior against the advisory). Two additions so this fully closes #601, plus one housekeeping ask:

1. **`plugins/protect-mcp/test/run-tests.sh`** still pins `protect-mcp@0.5.5` in 6 places and invokes a `keygen` subcommand that doesn't exist in 0.7.0 — please update the pins and swap `keygen` for the 0.7.0 equivalent.
2. **`plugins/review-agent-governance/skills/review-agent-setup/SKILL.md`** still shows `protect-mcp@0.5.5` in the copy-paste hook examples (~lines 162 and 171) — please bump those too. A repo-wide grep for `0.5.5` is worth a final pass.
3. Since you maintain protect-mcp upstream (and originally authored these plugins), please add a one-line disclosure of that relationship to the PR description — we just added a disclosure policy to CONTRIBUTING.md (#611), and it helps future reviewers.

CI is already green here; we'll re-check after you push.

tomjwxf:
Done, all three items:

1. **run-tests.sh**: pins bumped in all 6 places; `keygen` (removed in 0.7.0) swapped for `init --dir`. While fixing it I found the suite had deeper version drift: receipts are now read from `receipts.jsonl` (0.7.0+ appends one JSON per line), the expected schema is updated to the v2 envelope, and `fixtures/test-policy.cedar` is rewritten in the entity shape protect-mcp actually evaluates (`action == Action::"MCP::Tool::call"`, the tool as resource, input at `context.input`). The old `Action::"Read"` shape only appeared to work under 0.5.5's broken evaluator. Suite is 8/8 locally.
2. **SKILL.md** copy-paste examples bumped in both plugins. The remaining `0.5.5` strings under `docs/superpowers/` are dated investigation records where 0.5.5 is historically accurate, so I left those as written.
3. **Disclosure** added to the PR description per the #611 policy.

One deliberate deviation, flagged in the description: the pin is **0.7.4** rather than 0.7.0, because `evaluate` only passes tool input through to Cedar as `context.input` from 0.7.4, which the command-scoping test policy (and any real command-scoping policy) needs. 0.7.4 also carries the per-host `--format` hook adapters.

PR #622, opened 2026-07-14
Add software graph analysis skill

0xsarwagya:
## Summary\n- adds a developer-essentials skill for Ontoly Software Graph analysis\n- updates developer-essentials plugin metadata and docs counts\n- regenerates committed Cursor marketplace artifacts\n\n## Validation\n- make generate-all\n- make validate\n- make garden (0 errors; existing SKILL_OVER_CODEX_CAP warnings only)\n- git diff --check

### DIFF
--- a/.cursor-plugin/marketplace.json
+++ b/.cursor-plugin/marketplace.json
@@ -388,7 +388,7 @@
       "name": "developer-essentials",
       "source": "./plugins/developer-essentials",
       "version": "1.0.4",
-      "description": "Essential developer skills including Git workflows, SQL optimization, error handling, code review, E2E testing, authentication, debugging, and monorepo management",
+      "description": "Essential developer skills including Git workflows, SQL optimization, error handling, code review, E2E testing, authentication, debugging, monorepo management, and software graph analysis",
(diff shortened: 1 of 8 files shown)


wshobson:
Thanks for the contribution. The skill is well formed, and I can see you ran the validation targets before submitting.

I'm closing the PR, though, and I want to explain why. developer-essentials is a core plugin, and I keep core plugins vendor-neutral. The skill you wrote is specific to one product, Ontoly. It tells agents to prefer Ontoly's graph over reading the source and to run the `ontoly` CLI, so anyone who installs developer-essentials would get the Ontoly instructions by default. A skill that steers agents toward one vendor's tool should be something users choose to install on its own. In this marketplace it belongs in its own plugin, and hermes-tweet, runapi-mcp, and social-publishing all landed that way.

If you'd like to resubmit the skill as a standalone plugin, I'm happy to review it. A resubmission would need to be rebased on current main, since the counts in README.md and the docs have moved and the PR now conflicts, and it would need `make generate-all` run so the committed artifacts stay in sync.

Thanks again for taking the time.

PR #641, opened 2026-07-28
feat(opencode): register MiniMax provider and model IDs

octo-patch:
Reason: OpenCode model resolution rewrites explicit MiniMax model IDs to the inherit default because the provider/model registry lacks them.

Adds a third-party provider registry to `tools/adapters/capabilities.py` with the MiniMax provider, both regional endpoints (global and CN), and both native model IDs carrying context window, pricing, input modalities, and thinking modes sourced from the provider's published configuration.

`resolve_model()` now consults this registry for harnesses that emit provider-prefixed model IDs (detected via a `/` in their `inherit` alias, e.g. OpenCode). An explicit native model ID resolves to `<provider_id>/<model_id>` with no warning instead of being silently rewritten to the `inherit` fallback. Harnesses that use bare model IDs keep their existing alias/fallback behavior.

Checks:
- `uv run --project plugins/plugin-eval pytest -q tools/tests/test_adapters.py` — 70 passed.
- `uv run --project plugins/plugin-eval ruff check tools/adapters/capabilities.py` — all checks passed.
- `uv run --project plugins/plugin-eval python tools/generate.py --harness opencode --all` — no committed drift (output is gitignored).

### DIFF
--- a/tools/adapters/capabilities.py
+++ b/tools/adapters/capabilities.py
@@ -298,6 +298,101 @@ MODEL_ALIASES: dict[str, dict[str, str]] = {
 }
 
 
+@dataclass(frozen=True)
+class ProviderRegion:
+    """One regional endpoint set for a third-party model provider."""
+
+    region: str
+    openai_base_url: str
+    anthropic_base_url: str
+    docs_root: str
+
+
+@dataclass(frozen=True)
+class ProviderModel:
+    """One native model offered by a third-party provider."""
+
+    model_id: str
+    context_window: int
+    pricing_usd_per_million_tokens: dict[str, float | None]
+    input_modalities: tuple[str, ...]
+    thinking: tuple[str, ...]
+
+
+@dataclass(frozen=True)
+class ProviderEntry:
+    """A third-party provider with regional endpoints and native model IDs.
+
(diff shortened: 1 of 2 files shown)


wshobson:
Keeping this open, unmerged, until there is a demonstrated need. Nothing in the repo sets a MiniMax model ID, so the new code path is unreachable, and `resolve_model()` never reads most of the registry (pricing, modalities, endpoints). If a concrete use case appears, a minimal alias entry is the right size. Please also address the CodeRabbit findings.

wshobson:
Declining this one. Nothing in the repo references MiniMax anywhere — the adapters exist to ship this marketplace's own content to harnesses, so a provider registry nothing here uses is speculative. CONTRIBUTING asks for an issue with a concrete use case before that kind of addition.

Setting policy aside, the shape is top-heavy too. There's roughly 100 lines of endpoints, pricing, context windows and modality data, and `resolve_model()` only ever reads `model_id -> provider_id`. The `regions` field isn't read at all, and pricing in a static table starts rotting the day it lands.

If you have a real workflow that's actually blocked by this, open an issue describing it and we can look at the four lines it would take.

Issue #647, opened 2026-08-03
[NEW] attester-verify: existence-oracle checks before install


maminihds:
## Component proposal

**Name:** attester-verify (plugin containing one skill)

**What it does:** Teaches the agent to verify PyPI and npm package and symbol names against the attester.dev existence oracle before installing or importing them. One free keyless HTTP call per check; closest-match guidance on misses; quota exhaustion and offline treated as skip-and-continue, never as a blocker. It also covers pinning a remote MCP server's tool manifest so drift can be checked later.

**Why it belongs:** Models invent plausible package names: a USENIX Security 2025 study measured 5.2% to 21.7% of suggested package names as nonexistent, and attackers register those names ("slopsquatting"). Agents cannot self-check this failure mode. An oracle answering from real published artifacts (PyPI wheels, npm tarballs) closes it with one rule that works identically across all six harnesses.

**Cost to users:** none. The skill uses the free keyless tier: 25 calls per day per client IP, no account, no API key, no signup.

**Ownership disclosure:** I build and operate attester.dev (github.com/maminihds), the API this plugin wraps. The free tier is the default path and there is no paid funnel. The same disclosure and free-tier shape was accepted in github/awesome-copilot#2387 and xpaysh/awesome-x402#940.

**Readiness:** branch prepared with the plugin, marketplace entry, and regenerated harness registries. `make validate STRICT=1` clean; `make test` 487 passed; `make smoke-test` 5 passed / 3 skipped (local CLI absence per Makefile design); garden shows only pre-existing warnings in unrelated plugins.

wshobson:
Closing along with the direction on PR #648. CONTRIBUTING now has an [external and vendor plugin policy](https://github.com/wshobson/agents/blob/main/CONTRIBUTING.md#external-and-vendor-plugins); this proposal fails it on two points: a metered maintainer-operated API on the default path, and routing package names and MCP server URLs through a third-party service when the agent can query PyPI/npm directly. A skill that teaches direct registry verification (no intermediary) would be welcome.

PR #648, opened 2026-08-03
Add attester-verify plugin: existence-oracle checks before install

maminihds:
References: #647

## What this adds

`plugins/attester-verify`: one skill (`attester-verify`) that teaches the agent to verify PyPI and npm package and symbol names against the attester.dev existence oracle before installing or importing them, plus pinning of remote MCP server manifests for later drift checks. Ships to all six harnesses through the standard pipeline; `make generate-all` output is committed.

## Why

Models invent plausible package names: a USENIX Security 2025 study measured 5.2% to 21.7% of suggested package names as nonexistent, depending on model and ecosystem. Attackers register those names ("slopsquatting"). The oracle answers from real published artifacts (PyPI wheels, npm tarballs) and signs every negative, so "does not exist" is a checkable fact rather than a model opinion. This is a failure mode models cannot self-check, which is the gap a rules-based skill closes well.

## Ownership disclosure (required)

I build and operate attester.dev, the API this plugin wraps (github.com/maminihds). The plugin uses only the free keyless tier: 25 calls per day per client IP, no account, no API key, no signup. There is no funnel: the skill tells the agent to continue conservatively when the quota is spent, and the paid tier appears once in references/details.md as reference material, never as a call to action. The same disclosure and free-tier shape was accepted in github/awesome-copilot#2387 and xpaysh/awesome-x402#940.

## Validation

- `make generate-all`: 732 files written, 0 errors; registries for all five compiled harnesses committed.
- `make validate STRICT=1`: OK, no issues across 5 harnesses.
- `make garden STRICT=1`: 0 errors for this plugin. The run exits 1 on 10 pre-existing warnings (SKILL_OVER_CODEX_CAP) in ship-mate, signed-audit-trails, and startup-business-analyst; verified identical on a clean baseline without this change.
- `make test`: 487 passed, 3 skipped.
- `make smoke-test`: 5 passed, 3 skipped (CLI binaries not installed locally; skips per the Makefile design; CI installs them).
- Code-quality checks (ruff/ty on tools/ and src/plugin_eval/): findings only in files this PR does not touch; the change is Markdown and JSON only.

## Notes

- Lifecycle hooks are not portable across the six harnesses (per docs/harnesses.md), so the enforcement variant lives in a separate MIT repo (github.com/maminihds/attester-import-check: pre-commit hook, Claude Code hook, GitHub Action) and is linked from the skill's references file rather than bundled.

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -93,6 +93,19 @@
       "category": "Coding"
     },
     {
+      "name": "attester-verify",
+      "description": "Verify PyPI and npm package and symbol names against the attester.dev existence oracle before installing or importing, catching hallucinated dependencies",
+      "source": {
+        "source": "local",
+        "path": "./plugins/attester-verify"
+      },
+      "policy": {
+        "installation": "AVAILABLE",
+        "authentication": "ON_USE"
+      },
+      "category": "Coding"
+    },
+    {
       "name": "backend-api-security",
       "description": "API security hardening, authentication implementation, authorization patterns, rate limiting, and input validation",
       "source": {
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -1313,6 +1313,27 @@
         "deck-generation",
         "slide-design"
       ]
+    },
+    {
+      "name": "attester-verify",
+      "source": "./plugins/attester-verify",
+      "description": "Verify PyPI and npm package and symbol names against the attester.dev existence oracle before installing or importing, catching hallucinated dependencies with a free keyless tier (25 checks/day per IP).",
+      "version": "1.0.0",
+      "author": {
+        "name": "attester.dev",
+        "url": "https://github.com/maminihds"
+      },
+      "homepage": "https://attester.dev",
+      "license": "MIT",
+      "category": "security",
+      "keywords": [
+        "package-verification",
+        "hallucination",
+        "supply-chain",
+        "pypi",
+        "npm",
+        "slopsquatting"
+      ]
     }
   ]
 }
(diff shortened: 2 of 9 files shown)


wshobson:
Keeping this open rather than merging or closing. Open concerns:

- The skill routes package names and MCP server URLs through attester.dev. The agent can query PyPI and npm directly, with no quota and no third party.
- The 8 CodeRabbit findings are unaddressed, and the marketplace counts were never updated.
- The repo needs a written policy for plugins that wrap a maintainer-operated metered API before taking this on.

Respond to the review if you want to move it forward.

wshobson:
This is one of the more careful submissions I've had. The disclosure is thorough, the fail-open behavior on quota and network errors is exactly right, and the allowlist section for `yaml` / `PIL` / `cv2` shows you actually used the thing. I'm still going to decline, and the reason is policy rather than craft.

CONTRIBUTING picked up an [external and vendor plugin section](https://github.com/wshobson/agents/blob/main/CONTRIBUTING.md#external-and-vendor-plugins) on Aug 15, after you opened this. Two parts apply. A free daily quota with a paid tier behind it counts as a funnel whether or not it's disclosed. And a skill shouldn't route workspace data to a third-party endpoint when a first-party answer exists — package existence comes straight from `pypi.org/pypi/<name>/json` and `registry.npmjs.org/<name>`, free and unmetered, so I can't justify the oracle hop for every candidate name and every approved MCP server URL.

One small correction while I'm here: the PR says the paid tier appears once in `references/details.md`. It's also in the SKILL.md body (`/v2/mcp/check` at $0.01), and details.md carries the full price table plus the x402 rails. Not a big deal, but the disclosure undersells it slightly.

For what it's worth, the signed negatives and artifact proofs are a genuine difference from a plain registry lookup, and that's the strongest version of the argument for this plugin. It just doesn't outweigh the routing question for a catalog entry.

Sorry the policy landed mid-flight on this one.

PR #662, opened 2026-08-17
fix(plugin-eval): resolve nested skill references

### LINKED ISSUE #661: [BUG] DEAD_CROSS_REF false-positives for skills using a sub-skills/ subdirectory layout
### Preliminary Checks

- [x] I have read the Code of Conduct
- [x] I have searched existing issues to ensure this is not a duplicate

### Harness

Adapter framework / tooling (not a specific harness)

### Harness version

Upstream `main` at `d6837ae274c2cd817acad3fb98f193a4390a4c3e`.

### Affected plugin / agent / skill / command

`plugins/plugin-eval/src/plugin_eval/parser.py` (cross-ref extraction) and `plugins/plugin-eval/src/plugin_eval/layers/static.py` (`DEAD_CROSS_REF` resolution)

### Bug description

`parser.py`'s cross-reference extraction has no path-boundary guard, and `static.py`'s resolution only checks sibling directories — together this produces false-positive `DEAD_CROSS_REF` anti-pattern flags (each -0.05 severity) for any skill that legitimately organizes helper skills under a `sub-skills/` subdirectory, a pattern already in use by at least one real-world skill (a corrections-loop skill with 5 sub-skills, external to this repo).

**Bug 1 — unguarded regex** (`parser.py`):

```python
cross_refs = re.findall(r"(?:skill|skills)/([a-z0-9-]+)", body)
```

This matches the literal substring `skills/` anywhere in the body text, including inside `sub-skills/` — `sub-skills/intent-extractor` contains the substring `skills/intent-extractor`, so the regex wrongly extracts `intent-extractor` as a top-level cross-reference. There's no word/path-boundary check before `skill`/`skills`, so any `sub-skills/<name>` reference is misparsed as `skills/<name>`.

**Bug 2 — sibling-only resolution** (`static.py`, `DEAD_CROSS_REF` check):

```python
skill_parent = skill.path.parent  # skills/ dir
for ref in skill.cross_references:
    ref_path = skill_parent / ref
    if not ref_path.exists():
        # flagged as DEAD_CROSS_REF
```

Even for a correctly-extracted `sub-skills/<name>` reference, this only checks whether `<name>` exists as a *sibling* of the skill itself (`skill_parent / ref`), never `skill.path / "sub-skills" / ref` (nested inside the skill's own directory). So a skill's genuine, working references to its own sub-skills are unresolvable by design, independent of Bug 1.

### Steps to reproduce

```bash
cd plugins/plugin-eval
uv run python - <<'PY'
import re
from pathlib import Path

body = """See `sub-skills/intent-counter-example-checker/SKILL.md` for the full procedure.
- `.skills/correct/sub-skills/intent-extractor/SKILL.md`
- `.skills/correct/sub-skills/intent-dedup/SKILL.md`
- `.skills/correct/sub-skills/intent-regression-eval/SKILL.md`
- `.skills/correct/sub-skills/intent-applier/SKILL.md`"""

print(re.findall(r"(?:skill|skills)/([a-z0-9-]+)", body))
PY
```

Output:

```
['intent-counter-example-checker', 'correct', 'intent-extractor', 'correct',
 'intent-dedup', 'correct', 'intent-regression-eval', 'correct', 'intent-applier']
```

`correct` resolves fine (it's the skill's own directory — a sibling of itself, per the existing resolution logic), so it isn't flagged. The other 5 extracted names don't exist as siblings of the skill (they only exist nested under the skill's own `sub-skills/`), so a live `plugin-eval score <skill-dir> --depth standard` run against a real skill using this layout reports exactly 5 `DEAD_CROSS_REF` anti-patterns — a flat -25% penalty on the composite score — for correct, working cross-references.

### Expected behavior

1. The extraction regex should not match `skills/` when it's actually part of `sub-skills/` (e.g. require a path-segment boundary that isn't preceded by a word character or hyphen, such as `(?<![\w-])(?:skill|skills)/`).
2. Even where a genuine `sub-skills/<name>` reference is intended, the resolver should check `skill.path / "sub-skills" / name` (relative to the skill's own directory) in addition to `skill_parent / name` (sibling), so skills that organize helper skills under `sub-skills/` aren't penalized for a supported, real-world layout.

### Actual behavior

Any skill using a `sub-skills/<name>/` subdirectory layout gets a false `DEAD_CROSS_REF` flag for every such reference, incurring an undeserved anti-pattern penalty on its composite score.

### Additional context

Found via a live `plugin-eval score --depth standard` run against a real skill (external to this repo) that organizes 5 helper skills under `sub-skills/`, cross-checked by reading `parser.py`/`static.py` and independently reproducing the regex behavior shown above against the skill's actual `SKILL.md` text. Distinct from #591, #651, and the `model_usage` reporting bug filed separately — none of those touch cross-reference extraction or resolution.

uuzzrm:
## Summary

- Preserve nested `sub-skills/...` paths when parsing skill and agent cross-references.
- Resolve nested skill references relative to the current skill directory so valid helpers are not penalized.

## Scope

- [ ] Per-harness setup or docs (GEMINI.md / docs/harnesses.md)
- [ ] CI / build / release
- [ ] Other

## Affected harnesses

- [ ] Claude Code
- [ ] OpenAI Codex CLI
- [ ] Cursor
- [ ] OpenCode
- [ ] Gemini CLI

## Test plan

- [x] `make validate STRICT=1` equivalent: `uv run --project plugins/plugin-eval python -X utf8 tools/validate_generated.py --strict`
- [x] `make garden` equivalent: `uv run --project plugins/plugin-eval python -X utf8 tools/doc_gardener.py --strict` (0 errors, 10 pre-existing warnings)
- [ ] `make test` equivalent on Windows: 458 passed, 10 failed because Codex is unavailable and symlink creation is not permitted on this host
- [ ] `make smoke-test` was not fully green on this Windows host for the same environment constraints
- [x] `uv run pytest` in `plugins/plugin-eval/` (105 passed, 2 skipped)
- [x] Ran ruff + ty in `plugins/plugin-eval/`
- [ ] Spot-checked behavior in a real harness

N/A — pure tooling change.

## Related issue(s)

Closes #661

### DIFF
--- a/plugins/plugin-eval/src/plugin_eval/layers/static.py
+++ b/plugins/plugin-eval/src/plugin_eval/layers/static.py
@@ -233,6 +233,8 @@ class StaticAnalyzer:
         skill_parent = skill.path.parent  # skills/ dir
         for ref in skill.cross_references:
             ref_path = skill_parent / ref
+            if not ref_path.exists() and ref.startswith("sub-skills/"):
+                ref_path = skill.path / ref
             if not ref_path.exists():
                 patterns.append(
                     AntiPattern(
--- a/plugins/plugin-eval/src/plugin_eval/parser.py
+++ b/plugins/plugin-eval/src/plugin_eval/parser.py
@@ -9,6 +9,10 @@ from pathlib import Path
 
 import yaml
 
+_CROSS_REFERENCE_PATTERN = re.compile(
+    r"(?<![\w-])((?:skill|skills|sub-skills)/[a-z0-9-]+(?:/[a-z0-9-]+)*)"
+)
+
 
 @dataclass
 class ParsedSkill:
@@ -93,7 +97,7 @@ def parse_skill(skill_dir: Path) -> ParsedSkill:
     must_pattern = re.compile(r"\b(MUST|NEVER|ALWAYS)\b")
     must_count = len(must_pattern.findall(content))
 
-    cross_refs = re.findall(r"(?:skill|skills)/([a-z0-9-]+)", body)
+    cross_refs = _extract_cross_references(body)
 
     return ParsedSkill(
         path=skill_dir,
@@ -134,7 +138,7 @@ def parse_agent(agent_path: Path) -> ParsedAgent:
     description = frontmatter.get("description", "")
     has_proactive = bool(re.search(r"use proactively", description, re.IGNORECASE))
 
-    skill_refs = re.findall(r"(?:skill|skills)/([a-z0-9-]+)", body)
+    skill_refs = _extract_cross_references(body)
 
     return ParsedAgent(
         path=agent_path,
@@ -196,3 +200,13 @@ def _split_frontmatter(content: str) -> tuple[dict, str]:
         frontmatter = {}
 
     return frontmatter, parts[2]
+
+
+def _extract_cross_references(body: str) -> list[str]:
+    """Extract skill paths while keeping references to nested skills intact."""
+    references = []
+    for reference in _CROSS_REFERENCE_PATTERN.findall(body):
+        if reference.startswith(("skill/", "skills/")):
+            reference = reference.split("/", 1)[1]
+        references.append(reference)
+    return references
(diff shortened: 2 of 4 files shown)


wshobson:
Thanks @uuzzrm — clean fix for the sub-skills false positives, and the test coverage is appreciated. Merging.

PR #665, opened 2026-08-18
docs(readme): restore star history chart

OctoBored:
The star history chart was removed in d72efd3 because the old embed stopped rendering due to GitHub's stargazer API restrictions. Those restrictions no longer apply to the chart's data source, so we can bring the chart back without an API token. It is restored at the same location it previously occupied in the README, right after the License section.

Fixes the broken star history visualization referenced in d72efd3.


wshobson:
Thanks @OctoBored for restoring the chart — merging now.

PR #670, opened 2026-08-19
Add eskill plugin — meta-skill to build top-tier Agent Skills

hedralab:
## What

Adds **eskill** as a plugin: a meta-skill (SKILL.md) that guides agents through building spec-compliant Agent Skills — interview first, market/insight research, one-page plan (Pyramid), numbered-file pipeline, eval loop, validator, commercialization checklist.

## Details

- plugins/eskill/ — plugin manifest (.claude-plugin/plugin.json) + full skill at skills/eskill/ (SKILL.md, references/, scripts/, template/, examples/, eval-results.json)
- Registered in .claude-plugin/marketplace.json (category: development)
- Regenerated all harness registries via tools/generate.py (codex, copilot, cursor, opencode, antigravity)
- tools/validate_generated.py passes: OK, no issues across 5 harness(es)

MIT-licensed, free (no paid funnel). Repo: https://github.com/hedralab/eskill

Issue #643, opened 2026-07-31
[BUG] Same-named agents duplicated across plugins with divergent content (9 agents flagged by garden)


sergiopadure:
### Preliminary Checks

- [x] I have read the Code of Conduct
- [x] I have searched existing issues to ensure this is not a duplicate
- [x] This report contains only technical information about a bug

### Harness

Claude Code

### Harness version

N/A — repo/docs consistency issue, harness-independent

### Affected plugin / agent / skill / command

README.md, docs/agents.md, and the duplicated agents listed below (security-auditor, django-pro, fastapi-pro, python-pro)

### Bug description

Found during an automated repo audit (LLM-assisted; every claim below re-verified by hand against a fresh shallow clone, 2026-07-31). Three related consistency bugs, likely sharing one root cause (§3):

**1. README counts drift from the tree.** README.md claims 203 agents / 175 skills / 109 commands. Actual: **204** agents, **180** skills, 109 commands (matches).

**2. docs/agents.md model column is stale.** The catalog lists `sonnet` for python-pro, django-pro and fastapi-pro, but every corresponding file's frontmatter says `model: opus` (checked all copies of each).

**3. Same-named agents duplicated across plugins WITH divergent content** — not byte-identical copies; they have drifted apart, so fixes land in one copy and miss the others:
- `security-auditor.md` in **5** plugins (security-compliance, security-scanning, comprehensive-review, full-stack-orchestration, backend-development) — **all five md5s differ**
- `django-pro.md` in api-scaffolding + python-development — md5s differ
- `fastapi-pro.md` in api-scaffolding + python-development — md5s differ

### Steps to reproduce

```bash
git clone --depth 1 https://github.com/wshobson/agents && cd agents
find plugins -path '*/agents/*.md' | wc -l          # 204, README says 203
find plugins -name 'SKILL.md' | wc -l               # 180, README says 175
grep -E 'python-pro|django-pro|fastapi-pro' docs/agents.md   # model column: sonnet
grep '^model:' plugins/*/agents/{python-pro,django-pro,fastapi-pro}.md 2>/dev/null  # opus
find plugins -name security-auditor.md | xargs md5sum        # 5 distinct hashes
```

### Expected behavior

README counts match the tree; docs/agents.md model tiers match frontmatter; same-named agents are either a single canonical copy or kept identical by CI.

### Actual behavior

As described: counts off by +1/+5, model column stale, 9 divergent copies across 3 agent names.

### Additional context

[...]

PR #522, opened 2026-05-09
docs: synchronize plugin/agent/skill counts (closes #517)

wshobson:
## Summary

Companion to 159ce68 (README.md + CLAUDE.md). Six docs still carried stale counts; updates them to the live numbers verified against `find plugins -mindepth 1 -maxdepth 1 -type d | wc -l` (80 local), `find plugins -path '*/agents/*.md' | wc -l` (185), `find plugins -path '*/skills/*/SKILL.md' | wc -l` (153), and the 100 commands count from CLAUDE.md.

| File | Before → After |
|---|---|
| `.claude-plugin/marketplace.json` `metadata.description` | 79/184/150 → 80/185/153 |
| `docs/plugins.md` (header + footer link) | 78 plugins (77 local) → 81 plugins (80 local); 150 → 153 skills |
| `docs/agents.md` (header) | 184 agents → 185 |
| `docs/agent-skills.md` (intro + spec section) | 150 skills / 37 plugins → 153 / 40 |
| `docs/architecture.md` (3 places) | 79 plugins / 99 agents / 150 skills + "all 99 agents accessible" → 81 / 185 / 153 |
| `docs/usage.md` (2 places) | 150 skills, 79 plugins → 153 skills, 80 plugins |

## Test plan

- [x] `grep -nE '\b(78|79|184|99|150)\b' README.md CLAUDE.md .claude-plugin/marketplace.json docs/*.md | grep -iE 'plugin|agent|skill|command|specialized|focused|component'` returns no rows.
- [x] `jq '.metadata.description' .claude-plugin/marketplace.json` shows the new description.
- [x] All edits are scoped — no unrelated changes.

Closes #517.

PR #676, opened 2026-08-22
feat(garden): count and agent-divergence checks, plus tool-rewrite tests

wshobson:
## Summary

Three follow-ups from the open PR and issue audit. Each one closes a gap that
came up while reviewing the queue.

**Stale doc counts.** `check_doc_counts` compares the plugin, agent, skill, and
command totals quoted in README.md and AGENTS.md against the live counts, at
error severity. Only the two canonical context files get scanned, because docs/
carries per-category subtotals that legitimately differ from the totals. Plugins
count marketplace entries rather than `plugins/` directories, so external
git-subdir entries are included the way the headline figure counts them. Nothing
caught this before, which is why #522, #613, #615, and #616 all existed as manual
sync PRs. Counts are correct on main today, so the check is currently silent.

**Diverged agent copies.** `check_agent_divergence` groups agent files by
filename and compares bodies with the plugin-namespaced frontmatter `name:`
normalized away. Without that normalization no two copies can ever hash alike,
which is why #643 read the duplication as total divergence. Measured properly,
30 filenames are duplicated across 65 redundant files, 19 of them are verbatim
copies with no content difference, and 11 have genuinely drifted. Plugins install
individually and the marketplace points at `./plugins/<name>`, so a plugin has to
contain every agent it offers. Verbatim copies are therefore expected and report
as info, and drifted bodies report as warning.

**Untested tool rewriting.** `strip_claude_tool_refs` rewrites Claude tool names
in Copilot command bodies and had no test coverage. #642 changed its output in
three ways while claiming no behavioral change, and nothing caught it. The new
tests assert exact output for both `tool_case` values, both prose forms, and all
nine mapped tools.

**Policy.** The commercial section of CONTRIBUTING covered funnels to paid
products but not a payload that sells something itself. #670 shipped a Telegram
sales bot, USDT payment verification, and a script granting GitHub collaborator
access on payment, so the new bullet rules that out directly.

## Scope

- [ ] CI / build / release
- [x] Other (CONTRIBUTING policy)

## Affected harnesses

- [ ] Claude Code
- [ ] OpenAI Codex CLI
- [ ] Cursor
- [ ] OpenCode
- [ ] Antigravity CLI

## Test plan

- [x] `make validate STRICT=1` — OK, no issues across 5 harnesses
- [x] `make garden` — exit 0. Adds 11 `AGENT_BODY_DIVERGENT` warnings and 19
      `AGENT_COPY_REDUNDANT` info findings alongside the 10 pre-existing
      `SKILL_OVER_CODEX_CAP` warnings. No new errors, so the CI gate still passes.
- [x] `pytest tools/tests/` — 105 passed for the two touched files, 20 of them new
- [x] ruff check, ruff format --check, and ty, all run from `plugins/plugin-eval/`
      the way CI does — all clean

`make test` fails three round-trip tests on my machine
(`test_opencode_skill_count_matches_source`,
`test_copilot_skill_count_matches_source`,
`test_copilot_command_entrypoints_exist_for_every_plugin`). I confirmed the same
three fail on a clean checkout with these commits stashed, because `.opencode/`
and `.copilot/` are gitignored and go stale locally. CI regenerates them.

## Notes for review

The two new checks are deliberately non-blocking except for stale counts. The
divergence check reports what exists today rather than demanding a refactor,
because the underlying question is still open.

**Open decision, not settled here.** Either generate the duplicated agents from a
shared source during `make generate-all`, or keep hand-maintained copies and rely
on this check to catch drift. Collapsing to single canonical copies would break
standalone plugin installs, so that option is off the table. #643 stays open for
the remaining choice.

## Related issue(s)

Refs #643

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,6 +32,10 @@ Full frontmatter conventions in [`docs/authoring.md`](docs/authoring.md).
 - If a plugin wraps a third-party API, package, or service that you own or
   maintain, disclose that relationship in the PR description and the plugin
   README.
+- Do not ship payment processing, license enforcement, or access-control
+  tooling in a plugin payload. Scripts that take payment, verify a transaction,
+  or grant and revoke access to a repository or service are out of scope,
+  whether they charge the user or help the user charge someone else.
 
 ## External and vendor plugins
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,10 +32,14 @@ Full frontmatter conventions in [`docs/authoring.md`](docs/authoring.md).
 - If a plugin wraps a third-party API, package, or service that you own or
   maintain, disclose that relationship in the PR description and the plugin
   README.
-- Do not ship payment processing, license enforcement, or access-control
-  tooling in a plugin payload. Scripts that take payment, verify a transaction,
-  or grant and revoke access to a repository or service are out of scope,
-  whether they charge the user or help the user charge someone else.
+- Plugin payloads must not contain runnable machinery for collecting payment or
+  gating access. A script that takes payment, verifies a transaction, or grants
+  and revokes access to a repository or service is out of scope, whether it
+  charges the installing user or helps the installing user charge someone else.
+  Teaching an agent to build payment, licensing, or access-control features in
+  the user's own application is a different thing and is welcome — the
+  `payment-processing` plugin is the reference example. The line is whether the
+  payload runs a commercial transaction itself.
 
 ## External and vendor plugins
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,10 +36,13 @@ Full frontmatter conventions in [`docs/authoring.md`](docs/authoring.md).
   gating access. A script that takes payment, verifies a transaction, or grants
   and revokes access to a repository or service is out of scope, whether it
   charges the installing user or helps the installing user charge someone else.
-  Teaching an agent to build payment, licensing, or access-control features in
-  the user's own application is a different thing and is welcome — the
-  `payment-processing` plugin is the reference example. The line is whether the
-  payload runs a commercial transaction itself.
+  Verifying a transaction, checking a licence or entitlement, and granting or
+  revoking access are each covered on their own, so splitting the steps across
+  tools does not get around the rule. Teaching an agent to build payment,
+  licensing, or access-control features in the user's own application is a
+  different thing and is welcome, and the `payment-processing` plugin is the
+  reference example of that. The line is whether the payload operates the
+  contributor's commercial relationship or only explains how to build one.
 
 ## External and vendor plugins
 
--- a/tools/check_agent_name_collisions.py
+++ b/tools/check_agent_name_collisions.py
@@ -84,8 +84,7 @@ def main() -> int:
         return 0
 
     print(
-        f"Found {duplicate_name_count} duplicate agent names across "
-        f"{colliding_file_count} files:"
+        f"Found {duplicate_name_count} duplicate agent names across {colliding_file_count} files:"
     )
     for name, paths in duplicates.items():
         print(f"\n{name} ({len(paths)} files)")
@@ -93,20 +92,14 @@ def main() -> int:
             print(f"  - {path.relative_to(root)}")
 
     failed = args.fail_on_duplicates
-    if (
-        args.max_duplicate_names is not None
-        and duplicate_name_count > args.max_duplicate_names
-    ):
+    if args.max_duplicate_names is not None and duplicate_name_count > args.max_duplicate_names:
         print(
             f"\nERROR: duplicate name count {duplicate_name_count} exceeds "
             f"baseline {args.max_duplicate_names}",
             file=sys.stderr,
         )
         failed = True
-    if (
-        args.max_colliding_files is not None
-        and colliding_file_count > args.max_colliding_files
-    ):
+    if args.max_colliding_files is not None and colliding_file_count > args.max_colliding_files:
         print(
             f"\nERROR: colliding file count {colliding_file_count} exceeds "
             f"baseline {args.max_colliding_files}",
--- a/tools/doc_gardener.py
+++ b/tools/doc_gardener.py
@@ -69,7 +69,7 @@ COUNT_NOUN_ALIASES = {
 # agent can never hash alike. Normalize it away before comparing bodies, scoped to
 # the opening frontmatter block so a `name:` line in the body still counts as content.
 AGENT_NAME_LINE_RE = re.compile(r"^name:.*$", re.MULTILINE)
-AGENT_FRONTMATTER_RE = re.compile(r"\A(---\n)(.*?)(\n---\n)", re.DOTALL)
+AGENT_FRONTMATTER_RE = re.compile(r"\A(---\r?\n)(.*?)(\r?\n---(?:\r?\n|\Z))", re.DOTALL)
 
 
 # ── Findings ─────────────────────────────────────────────────────────────────
@@ -431,9 +431,14 @@ def actual_counts() -> dict[str, int | None]:
     plugins: int | None = None
     if MARKETPLACE_JSON.is_file():
         try:
-            plugins = len(json.loads(MARKETPLACE_JSON.read_text()).get("plugins", []))
+            manifest = json.loads(MARKETPLACE_JSON.read_text())
         except json.JSONDecodeError:
-            plugins = None  # check_marketplace_consistency reports the parse error
+            manifest = None  # check_marketplace_consistency reports the parse error
+        # Valid JSON of the wrong shape is still an unknown count, not a crash.
+        if isinstance(manifest, dict):
+            entries = manifest.get("plugins")
+            if isinstance(entries, list):
+                plugins = len(entries)
     return {
         "plugins": plugins,
         "agents": len(list(PLUGINS_DIR.glob("*/agents/*.md"))),
--- a/tools/doc_gardener.py
+++ b/tools/doc_gardener.py
@@ -65,11 +65,10 @@ COUNT_NOUN_ALIASES = {
     "command": "commands",
 }
 
-# An agent's frontmatter `name:` is plugin-namespaced, so two copies of the same
-# agent can never hash alike. Normalize it away before comparing bodies, scoped to
-# the opening frontmatter block so a `name:` line in the body still counts as content.
-AGENT_NAME_LINE_RE = re.compile(r"^name:.*$", re.MULTILINE)
-AGENT_FRONTMATTER_RE = re.compile(r"\A(---\r?\n)(.*?)(\r?\n---(?:\r?\n|\Z))", re.DOTALL)
+# An agent's frontmatter `name:` is plugin-namespaced, so two copies of the same agent
+# can never compare equal on raw text. It is dropped before comparing. A byte-order
+# mark or leading blank line would otherwise hide the frontmatter from the parser.
+BOM = "\ufeff"
 
 
 # ── Findings ─────────────────────────────────────────────────────────────────
@@ -406,17 +405,18 @@ def check_marketplace_consistency(report: Report) -> None:
 
 
 def normalized_agent_text(text: str) -> str:
-    """Blank the plugin-namespaced frontmatter `name:` so sibling copies compare equal.
+    """Render an agent as its frontmatter fields minus `name`, plus its body.
 
-    Only the opening frontmatter block is touched. A `name:` line in the body is real
-    content and must still count toward a divergence.
+    Uses the same frontmatter parser the adapters use, so a copy is judged on its
+    fields and body rather than on exact delimiter formatting. That keeps CRLF files,
+    a closing `---` at end of file, and a trailing space after a delimiter from
+    reading as drift. A `name:` line in the body is body content and still counts.
     """
-    match = AGENT_FRONTMATTER_RE.match(text)
-    if not match:
-        return text
-    opener, frontmatter, closer = match.groups()
-    frontmatter = AGENT_NAME_LINE_RE.sub("name:", frontmatter, count=1)
-    return f"{opener}{frontmatter}{closer}{text[match.end() :]}"
+    text = text.lstrip(BOM).lstrip()
+    fields, body = parse_frontmatter(text)
+    fields.pop("name", None)
+    rendered = "\n".join(f"{key}: {fields[key]!r}" for key in sorted(fields))
+    return f"{rendered}\n---\n{body.strip()}"
 
 
 def actual_counts() -> dict[str, int | None]:
--- a/tools/doc_gardener.py
+++ b/tools/doc_gardener.py
@@ -55,8 +55,10 @@ COUNT_RE = re.compile(r"\b(\d+)\s+(plugins|agents|subagents|skills|commands)\b")
 COUNT_NOUN_ALIASES = {"subagents": "agents"}
 
 # An agent's frontmatter `name:` is plugin-namespaced, so two copies of the same
-# agent can never hash alike. Normalize it away before comparing bodies.
+# agent can never hash alike. Normalize it away before comparing bodies, scoped to
+# the opening frontmatter block so a `name:` line in the body still counts as content.
 AGENT_NAME_LINE_RE = re.compile(r"^name:.*$", re.MULTILINE)
+AGENT_FRONTMATTER_RE = re.compile(r"\A(---\n)(.*?)(\n---\n)", re.DOTALL)
 
 
 # ── Findings ─────────────────────────────────────────────────────────────────
@@ -392,6 +394,20 @@ def check_marketplace_consistency(report: Report) -> None:
             )
 
 
+def normalized_agent_text(text: str) -> str:
+    """Blank the plugin-namespaced frontmatter `name:` so sibling copies compare equal.
+
+    Only the opening frontmatter block is touched. A `name:` line in the body is real
+    content and must still count toward a divergence.
+    """
+    match = AGENT_FRONTMATTER_RE.match(text)
+    if not match:
+        return text
+    opener, frontmatter, closer = match.groups()
+    frontmatter = AGENT_NAME_LINE_RE.sub("name:", frontmatter, count=1)
+    return f"{opener}{frontmatter}{closer}{text[match.end() :]}"
+
+
 def actual_counts() -> dict[str, int]:
     """Live component totals, counted the same way the adapters discover them.
 
@@ -451,7 +467,7 @@ def check_agent_divergence(report: Report) -> None:
             continue
         bodies: dict[str, list[Path]] = defaultdict(list)
         for path in paths:
-            normalized = AGENT_NAME_LINE_RE.sub("name:", path.read_text(encoding="utf-8"), count=1)
+            normalized = normalized_agent_text(path.read_text(encoding="utf-8"))
             bodies[hashlib.md5(normalized.encode("utf-8")).hexdigest()].append(path)
 
         if len(bodies) > 1:
--- a/tools/doc_gardener.py
+++ b/tools/doc_gardener.py
@@ -467,6 +467,20 @@ def check_marketplace_consistency(report: Report) -> None:
             )
 
 
+def canonical_frontmatter_value(value: object) -> object:
+    """Order-independent form of a parsed frontmatter value.
+
+    Mapping key order carries no meaning, and `repr` preserves insertion order, so two
+    agents with the same nested fields written in a different order would otherwise
+    hash differently. List order is left alone because it is meaningful.
+    """
+    if isinstance(value, dict):
+        return {key: canonical_frontmatter_value(value[key]) for key in sorted(value)}
+    if isinstance(value, list):
+        return [canonical_frontmatter_value(item) for item in value]
+    return value
+
+
 def normalized_agent_text(text: str) -> str:
     """Render an agent as its frontmatter fields minus `name`, plus its body.
 
@@ -483,7 +497,9 @@ def normalized_agent_text(text: str) -> str:
         text = trimmed
     fields, body = parse_frontmatter(text)
     fields.pop("name", None)
-    rendered = "\n".join(f"{key}: {fields[key]!r}" for key in sorted(fields))
+    rendered = "\n".join(
+        f"{key}: {canonical_frontmatter_value(fields[key])!r}" for key in sorted(fields)
+    )
     # Line endings are formatting, so a CRLF copy must match its LF twin.
     # parse_frontmatter strips leading "\n" but leaves the "\r" behind it.
     # Strip newlines only, never spaces: leading indentation is content, so an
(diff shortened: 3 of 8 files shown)


wshobson:
All three findings addressed in `9056298`. One of them had a wrong premise, so here's what I checked.

**One-digit counts (CodeRabbit).** Valid, applied as suggested. The two-digit floor was defensive against per-category subtotals like `### Development (6 plugins)`, but the check only scans README.md and AGENTS.md and those subtotals live under `docs/`. So the floor bought nothing and created a blind spot. `COUNT_RE` now uses `\d+`, and there's a test for a one-digit mismatch.

**`subagents` not matched (Codex).** Valid. AGENTS.md line 65 says `202 subagents`, and `\s+agents\b` can't match it because the word starts with `sub`. Added `subagents` as an alias for the agent total, with tests for both the matching and mismatching case. Good catch, this is the mention most likely to be missed by hand.

**Info findings and strict mode (Codex).** The conclusion is right and I applied it, but the stated cause isn't. `make garden STRICT=1` was already exiting 1 on main before this PR, because of the 10 pre-existing `SKILL_OVER_CODEX_CAP` warnings, and line 30 fails strict on warnings as well as info. I verified by running `--strict` on `main` with these commits stashed, and it exits 1 there too. So the contributor gate was already red and this PR didn't turn it.

The underlying point still holds on its own, so I acted on it. A copy that matches byte for byte once the namespaced `name:` is normalized is expected rather than drift, and 19 info findings only added noise to a check whose job is finding drift. `check_agent_divergence` now stays silent on verbatim copies and only warns when bodies actually differ. Garden output drops from 30 new findings to 11.

Worth flagging separately: strict mode being permanently red on main is a real problem, since CONTRIBUTING tells contributors to run it. Two recent PRs noted the same thing in their test plans. That's a follow-up rather than something to fix here.

Gates after the change, all run the way CI runs them. Ruff check, ruff format and ty are clean from `plugins/plugin-eval/`. `pytest tools/tests/` gives 389 passed with 26 in the gardener file. `make garden` exits 0. The three `test_round_trip` failures are the pre-existing stale `.opencode/` and `.copilot/` trees, identical on a clean checkout.

wshobson:
@coderabbitai full review

Incremental reviews are disabled, so `9056298` hasn't been looked at yet. All three findings from the first pass are addressed in that commit, with the reasoning in the comment above. CI is fully green and Claude's re-review passed.

wshobson:
Good catch on the ambiguity, fixed in `e0b1c5f`. I didn't take either of the two proposed remedies though, because neither fits what's actually in the repo.

I checked `plugins/payment-processing/` before changing anything. It's 11 files, all Markdown and JSON, with no executable scripts at all. Four skills covering Stripe integration, PayPal integration, subscription billing, and PCI compliance, plus one agent. The `requests.post` and `fetch` hits are illustrative code samples inside SKILL.md and details.md, and the keys in them are `sk_test_...` placeholders. It teaches an agent to build payment features in the user's own application. It's authored by the repo owner and MIT licensed.

So there's no policy conflict to resolve. Removing the plugin would delete legitimate domain content, and grandfathering it would imply it was ever in violation, which would also make the rule read as banning any future plugin about payments. Both remedies would make the policy worse.

The real problem is that my wording was sloppy. "Do not ship payment processing ... tooling in a plugin payload" reads as banning a plugin *about* payment processing, which was never the target. The target is #670's payload, which shipped a runnable Telegram sales bot, a Tronscan transaction verifier, and a `gh api` script granting repo collaborator access, all wired to the contributor's own USDT wallet.

The rule now turns on that distinction. A payload must not contain runnable machinery that collects payment or gates access, and teaching an agent to build those features in the user's own application stays welcome, with `payment-processing` named as the reference example. The deciding question is whether the payload runs a commercial transaction itself.

That should be a more useful rule than what I wrote first, so the finding was worth raising even though the suggested fixes weren't the way there.

markdownlint is clean on the file. No code changed in this commit, so the gardener and test results from the previous round still stand.

wshobson:
@coderabbitai full review

wshobson:
Correction to my comment above: the commit is `0b238d6`, not `e0b1c5f`. I typed the wrong hash.

wshobson:
@coderabbitai full review

The previous `full review` finished 7 seconds after I asked, which was the tail of the run started at 22:48 against `9056298`. Commit `0b238d6` landed at 22:51:54, so it hasn't been reviewed yet.

`0b238d6` rewrites the CONTRIBUTING bullet that your last finding was about. All 10 checks are green on it.

wshobson:
Applied in `691f25f`, though the premise needs one correction.

`tool_case` defaults to `"lower"` (`tools/adapters/base.py:474`), so the mapping loop was already exercising the lower path for all nine tools, not four. The four-mapping count describes the exact-output tests above it, where the sample sentence only mentions Read, Bash, Grep and Glob.

The change is still worth making, for a different reason than stated. The loop depended on the default rather than saying which case it meant, and normal mode was pinned for Read alone. Eight of the nine tools resolve to the same verb in both modes today, so a divergence in any of them would have gone unnoticed.

The loop now asserts both modes explicitly for all nine, with expected values verified against the current implementation rather than assumed. Only Read differs, `open` in lower and `read` in normal. I also added a small test pinning that the default is `lower`, since Copilot emission calls it both ways and the two call sites in `copilot.py` pass it explicitly while other callers would get the default.

Gates: 109 passed across the two test files, ruff check, ruff format and ty clean from `plugins/plugin-eval/`, garden exits 0. Test-only commit, so nothing else moved.

wshobson:
@coderabbitai full review

wshobson:
Applied in `4798d40`. This one was a genuine correctness gap, though it has no effect on the repo as it stands today.

I checked before changing anything. Across all 202 agent files, every one has a frontmatter `name:`, none has a `^name:` line anywhere in its body, and the first match is inside the frontmatter in every case. So `count=1` was landing on the right line 202 times out of 202, and the divergence result is identical before and after this commit at 11 divergent.

The gap is real for content the repo doesn't have yet. An agent whose frontmatter omitted `name:` would have had the first body `name:` line blanked instead, which is the opposite of what the function is for. Normalization is now scoped to the opening frontmatter block via `normalized_agent_text`, so a `name:` in the body stays content.

Three tests added: differing body `name:` lines still report `AGENT_BODY_DIVERGENT`, matching body `name:` lines with differing frontmatter names stay a verbatim copy, and a file with no frontmatter at all doesn't crash.

Gates: 393 passed in `tools/tests/`, 29 in the gardener file. ruff check, ruff format and ty clean from `plugins/plugin-eval/`. `make garden` exits 0, `validate STRICT=1` exits 0. The 3 `test_round_trip` failures are the pre-existing stale `.opencode/` and `.copilot/` trees, unchanged and identical on a clean checkout.

wshobson:
@coderabbitai full review

wshobson:
Applied in `127ad00`. I'd missed this thread earlier, since it was opened at 23:31:29 just ahead of the review body I read, and it's the one thing still holding the PR at `CHANGES_REQUESTED`.

`COUNT_RE` now matches singular forms and maps them to the same keys. Neither README.md nor AGENTS.md contains a singular count today, so the live result is unchanged and the counts check stays silent. Two tests added, one for a mismatched singular claim and one for a matching singular claim.

One tradeoff worth recording, since it's the reason the check is scoped the way it is. Matching singular nouns means prose like "each plugin ships 1 agent" would now read as a stale total and fail at error severity. That's tolerable here only because the check scans exactly two curated files that the maintainer writes. It would be unsafe to point this at `docs/`. I put that reasoning in a comment above the pattern so the constraint travels with the code.

Gates: 395 passed in `tools/tests/`, 31 in the gardener file. ruff check, ruff format and ty clean from `plugins/plugin-eval/`. `make garden` exits 0. The 3 `test_round_trip` failures are the pre-existing stale `.opencode/` and `.copilot/` trees, unchanged.

That should be every open thread. The other five are resolved.

wshobson:
@coderabbitai full review

wshobson:
Valid, and the worst of the findings so far. Fixed in `ecd10c8`.

A JSON syntax error in `marketplace.json` set the plugin count to 0, so `check_doc_counts` flagged every documented plugin mention as stale and told the reader to write 0. Six wrong errors at error severity, with the one real error from `check_marketplace_consistency` buried underneath. My own comment on that line said the parse error was reported elsewhere, which is true and beside the point, since the bogus advice still went out.

The count is now `None` when the manifest is missing or unparseable, meaning unknown rather than zero, and the comparison skips that key. Agent, skill and command counts are still checked, so a broken manifest no longer suppresses everything either. Two tests cover the unparseable and missing cases.

Gates: 397 passed in `tools/tests/`, 33 in the gardener file. ruff check, ruff format and ty clean from `plugins/plugin-eval/` (ty is happy with the `int | None` return). `make garden` exits 0. The 3 `test_round_trip` failures are the pre-existing stale `.opencode/` and `.copilot/` trees.

wshobson:
@coderabbitai full review

wshobson:
Both valid, both confirmed by running them rather than reading them. Fixed in `edeb74a`.

**CRLF and EOF frontmatter.** This one is a regression I introduced. Before `4798d40` the normalization was a `MULTILINE` `^name:` substitution over the whole file, which handles CRLF fine. Scoping it to the frontmatter block traded that away, because the new pattern required LF delimiters and a newline after the closing `---`. A CRLF agent, or one that ends at its frontmatter with no body, got no normalization at all, so two copies differing only in their namespaced name read as divergent. Reproduced both before the fix:

```
CRLF normalize equal?  False   (want True)
EOF-close equal?       False   (want True)
```

The pattern now takes `\r?\n` and allows the closing delimiter at EOF. Both cases pass. No agent in the repo uses CRLF today, so live output is unchanged at 11 divergent, but the check would have been wrong the first time someone committed one from Windows.

**Malformed manifest shape.** Also real, and it crashes rather than misreports:

```
[]                   AttributeError: 'list' object has no attribute 'get'
{"plugins": null}    TypeError: object of type 'NoneType' has no len()
```

Neither is a `JSONDecodeError`, so my handler didn't catch them and `--check counts` exited with a traceback. The shape is now validated before `len()`, and both yield an unknown count the same way a parse failure does.

Four regression tests: CRLF copies, frontmatter closing at EOF, list root, and null plugins.

Gates: 401 passed in `tools/tests/`, 37 in the gardener file. ruff check, ruff format and ty clean from `plugins/plugin-eval/`. `make garden` exits 0 with the same 11 warnings. The 3 `test_round_trip` failures are the pre-existing stale `.opencode/` and `.copilot/` trees.

wshobson:
@coderabbitai full review

wshobson:
Pushed `fe392c6` ahead of the next review, because rounds four and seven were both the same underlying problem and widening the regex a third time would only invite a fourth.

The root cause was that I hand-rolled a frontmatter matcher when `parse_frontmatter` was already imported into this module and is more tolerant than what I wrote. Agent comparison now goes through that shared parser and drops the namespaced `name` field, so a copy is judged on its frontmatter fields and body rather than on delimiter formatting. Both regexes are gone.

I checked the remaining cases in that class before pushing, and three more were broken that nobody had flagged yet:

```
BOM             normalizes equal? False
leading blank   normalizes equal? False
trailing space  normalizes equal? False
```

All three pass now. The local battery covers 16 shapes: CRLF, BOM, leading blank line, trailing space after a delimiter, closing `---` at EOF, no trailing newline, extra blank lines in the body, a `name:` line inside the body, block scalars and inline lists all compare equal when only the name differs; body text, model tier, an added field, a differing body `name:` line and files with no frontmatter all still register as drift.

Behavior on the real repo is unchanged, which was the thing I most wanted to confirm given the comparison basis moved from raw text to parsed fields. Still 11 divergent and 19 verbatim, with identical names and copy counts:

```
divergent: 11   verbatim: 19
  cloud-architect.md (4 copies / 2 versions)
  code-reviewer.md (7 copies / 2 versions)
  ...
  test-automator.md (6 copies / 3 versions)
```

Gates: 406 passed in `tools/tests/`, 42 in the gardener file. ruff check, ruff format and ty clean from `plugins/plugin-eval/`. `make garden` exits 0, `validate STRICT=1` exits 0.

wshobson:
@coderabbitai full review

wshobson:
`1a5cc3c` fixes one more, found by running a battery over `check_doc_counts` rather than by review.

`COUNT_RE` matched bare digit runs, so `1,234 agents` matched the `234` and reported "says 234 agents, actual is 34". A comma-formatted total anywhere in README.md or AGENTS.md would have failed CI at error severity with wrong advice, which is the same failure shape as the manifest-zero bug from an earlier round. The pattern now takes the whole number and strips separators, so 1,234 reads as 1234 and gets checked properly instead of misread.

The rest of the battery came back clean, and two results are worth recording as deliberate rather than accidental:

- Counts inside code fences **are** checked. That looked like a false positive until I read README.md line 22, where `# or any of 92 plugins` sits inside a bash fence and is a real total. There's now a test pinning it.
- `**12** plugins`, with the number bolded separately from the noun, is not matched. No file uses that form, and widening the pattern to allow markup between the number and the noun would cost more than it buys. Noted rather than fixed.

One thing I am deliberately not changing. A non-UTF-8 README raises `UnicodeDecodeError` and takes the gardener down. That is true of `check_dead_links` and `check_codex_skill_caps` too, since every check in this file uses a bare `read_text()`. Hardening only the two new checks would leave the module inconsistent and still crashable, so it belongs in its own change rather than here.

Gates: 409 passed in `tools/tests/`, 45 in the gardener file. ruff check, ruff format and ty clean from `plugins/plugin-eval/`. `make garden` exits 0 and the counts check is silent on the real repo.

wshobson:
@coderabbitai full review

wshobson:
@coderabbitai full review

wshobson:
All four addressed in `8a02f34`. Every one was valid, and the indentation finding caught a bug I introduced two commits ago.

**`body.strip()` masked real drift.** When I moved comparison onto `parse_frontmatter`, I stripped the body at both ends. Leading whitespace is content in Markdown, so an indented code block compared equal to the same text as plain prose:

```
indented vs not, equal? True   (want False)
```

Now `rstrip` only. A trailing newline is formatting; leading indentation is not. Two tests added, one for indentation as drift and one for matching indentation staying verbatim.

**Unreadable manifest wasn't reported.** `actual_counts` read `marketplace.json` directly, so a non-UTF-8 manifest silently became an unknown plugin count with no `UNREADABLE_FILE` finding. It now takes the `Report` and reads through `read_text_or_none` like every other read. My previous commit message claimed it had no `Report` to write to, which was a limitation I accepted rather than one that was forced. Passing it in was the better call.

**Dead `try/except OSError`** in `check_dead_links`, left behind when I wired in the shared reader. `read_text_or_none` already catches `OSError` and returns `None`, so the outer guard was unreachable. Thanks to the Claude bot for that one. Removed.

**CONTRIBUTING wording.** Correct catch, and the more serious of the four. My closing sentence narrowed the rule to running a transaction, which would have permitted exactly half of what #670 shipped: `grant_access.sh` consumes an external payment result and only flips repo access. Verification, entitlement checks and access changes are now each covered on their own, with a note that splitting the steps across tools doesn't get around it.

I re-ran the local battery, now 18 cases. All pass, including the new indentation pair, and real-repo output is unchanged at 11 divergent.

Gates: 414 passed in `tools/tests/`, 50 in the gardener file. ruff check, ruff format and ty clean. `make garden` exits 0, `validate STRICT=1` exits 0, markdownlint clean on CONTRIBUTING.

Separately, #677 now carries the pre-existing tooling issues found along the way, including the ruff scope gap that let four files drift out of format.

wshobson:
@coderabbitai full review

wshobson:
Both threads addressed, in `aaf76f6` and `5466963`. Reproduced everything first, and one of the two turned up a third problem.

**Leading whitespace without frontmatter.** Correct, and it is the same mistake as last round's `body.strip()` one level up. I stripped leading whitespace before deciding whether frontmatter existed, so two frontmatter-free agents differing only by indentation compared equal:

```
indented vs not (no frontmatter), equal? True   (want False)
```

Blank lines ahead of a frontmatter block are formatting. Ahead of anything else they are body content. Now conditional on the trimmed text starting with `---`, so the leading-blank-plus-frontmatter case still matches.

**Malformed plugin entries.** Also correct. A `plugins` list holding a non-object was counted as valid, so `{"plugins": [null, null]}` yielded 2 and would have reported "says 92 plugins, actual is 2" at error severity. The count is unknown now unless every entry is an object, matching how a parse failure and a wrong root shape are already handled.

**The third one, which nobody flagged.** Checking that input against the neighbouring check showed it crashes:

```
check_marketplace_consistency: CRASH AttributeError: 'NoneType' object has no attribute 'get'
```

A non-object root does the same. Both now report `MARKETPLACE_SHAPE` and continue. That crash predates this PR.

**A mistake on my side worth stating.** I pushed `aaf76f6` without reading the ty output, and it carried two `invalid-argument-type` errors: the new `isinstance` guard narrowed the entry from `Any` to `dict[Unknown, Unknown]`, so `.get("name")` no longer type-checked. `5466963` fixes it with a single cast after the guard. CI would have caught it, but I should have.

Three regression tests added: indentation without frontmatter, a malformed entry making the count unknown, and both bad manifest shapes reporting instead of raising.

Gates on `5466963`: 418 passed in `tools/tests/`, 54 in the gardener file. ruff check, ruff format and ty all clean. `make garden` exits 0.

wshobson:
@coderabbitai full review

wshobson:
All three in `3fd4455`. The CRLF one is the most useful finding of the whole review, because it exposed a flaw in how I was testing rather than just a line of code.

**CRLF against LF.** `parse_frontmatter` strips a leading `\n` but leaves the `\r` behind it, so a CRLF copy never matched its LF twin:

```
CRLF vs LF equal? False   (want True)
crlf body: "us'\n---\n\r\nBody text."
lf   body: "opus'\n---\nBody text."
```

My own CRLF test passed the whole time because it compared CRLF against CRLF, which succeeds without normalizing anything. The case that actually happens is one copy edited on Windows and the other on Unix, and I never wrote it. The test now mixes the two endings, and a second test proves that normalizing line endings does not also flatten a genuine body difference. Newlines are stripped, spaces never, so indentation stays content.

**Unhashable `name`.** Confirmed, `{"name": ["bad"], "source": {}}` raised `TypeError: unhashable type: 'list'` on `external_names.add(name)`. A non-string name now reports `MARKETPLACE_SHAPE` and the entry is skipped.

**The weak default-case test.** Fair, and I had reached for the easy assertion. `CopilotAdapter` passes `tool_case="lower"` at both call sites, so a test on the method default could stay green while generated commands regressed. Replaced with one that runs a command through `CopilotAdapter.emit_plugin` and asserts the emitted body says `` `open` ``, `` `shell` `` and `` `grep` `` with the original phrasings gone. That covers the path that would actually break.

Gates: 420 passed in `tools/tests/`, 56 in the gardener file. ruff check, ruff format and ty clean. `make garden` exits 0, output unchanged at 11 divergent.

wshobson:
@coderabbitai full review

wshobson:
Both in `1fcf75b`, and the second one turned up a flaw in the test I had just written.

**Closing-delimiter whitespace.** Confirmed:

```
closing "--- " vs "---" equal? False   (want True)
a body tail: "'\n---\n \nBody text."
b body tail: "us'\n---\nBody text."
```

`parse_frontmatter` leaves the spaces from a `--- ` closing line at the head of the body. My earlier trailing-space test put the space on the *opening* delimiter, which `parse_frontmatter` already tolerates, so it never exercised this. Leading whitespace-only lines are now treated as delimiter residue and dropped, while a line that starts with spaces followed by real text keeps its indentation. Six-case check after the change: closing-delimiter, CRLF vs LF, blank lines before body, indentation preserved, no-frontmatter indentation, and real drift all behave correctly.

**Per-file Copilot assertions.** Right, and worse than described. Asserting per file immediately failed, which exposed that my test had been asserting over *every* emitted markdown file, including `index.md`. That file is a plugin index with no command body, and the joined string had been hiding it. So the join was masking two things at once. The test now selects the two actual command artifacts, asserts each contains `` `open` ``, `` `shell` `` and `` `grep` ``, and rejects `` `read` `` in both.

Gates: 421 passed in `tools/tests/`, 57 in the gardener file. ruff check, ruff format and ty clean on this PR's CI scope. `make garden` exits 0, output unchanged at 11 divergent.

The two `B007` findings ruff reports over the full `tools/` tree are pre-existing in `yt-design-extractor.py`, outside this PR's CI scope, and fixed in #677.

wshobson:
@coderabbitai full review

wshobson:
Valid, fixed in `e7bd51c` with a lint follow-up in `8e61a7a`.

Confirmed the inconsistency before changing anything:

```
actual_counts plugins = 1        (counts it because it is a dict)
consistency kinds     = ['MARKETPLACE_SHAPE']   (rejects the same entry)
```

I took the second option you offered rather than the first. Copying the name rule into `actual_counts` would have fixed this instance and left the two readers free to drift apart again on the next rule. Both now call `marketplace_entry_problem`, which returns why an entry is unusable or `None` when it is fine. That also collapsed the two near-identical report blocks in the consistency check into one.

This is the same shape as the `parse_frontmatter` refactor earlier in the review: two places implementing one rule, fixed by having one place implement it.

Coverage is parametrized over the rule itself, eight cases including an optional name, an empty name, a list name, an integer name, and three non-object entries, plus a case proving a list-valued name leaves the count unknown.

**A process failure worth naming.** I pushed `e7bd51c` with an `I001` on the import block, having not read the ruff output before committing. That is the second time in this review, after the `ty` errors in `aaf76f6`. Both were visible on screen. `8e61a7a` fixes it, and I am running `make lint` as a gate before pushing rather than scanning output.

Gates on `8e61a7a`: 430 passed in `tools/tests/`, 66 in the gardener file. ruff check, ruff format and ty all clean on this PR's CI scope. `make garden` exits 0, `validate STRICT=1` exits 0, output unchanged at 11 divergent.

wshobson:
@coderabbitai full review

wshobson:
Valid, fixed in `ed42169`. Reproduced first:

```
parsed a: {'name': 'alpha-r', 'metadata': {'version': '1.0.0', 'author': 'me'}}
parsed b: {'name': 'beta-r',  'metadata': {'author': 'me', 'version': '1.0.0'}}
equal? False   (want True)
```

I sorted the top-level keys and stopped there, which left `repr` free to preserve insertion order inside any nested mapping. Keys are now sorted recursively.

One thing I deliberately did not do: sort lists. Mapping key order carries no meaning, but list order does. Reordering `tools: [Read, Write]` is a real change to an agent and should keep reporting as drift, so the canonicalizer recurses into lists without reordering them. Three tests pin the distinction: reordered nested keys are not drift, a changed nested value is, and a reordered list is.

No agent in the repo uses nested frontmatter today, so live output is unchanged at 11 divergent. Skills do use it, and nothing stops an agent from adding a `metadata:` block, so it was worth closing.

Gates, all run before pushing this time rather than after: 433 passed in `tools/tests/`, 69 in the gardener file. ruff check, ruff format and ty clean on this PR's CI scope. `make garden` exits 0, `validate STRICT=1` exits 0.

wshobson:
@coderabbitai full review

PR #677, opened 2026-08-23
chore(tooling): close the ruff scope gap and correct two stale docs

wshobson:
## Summary

Pre-existing issues found while working on #676. Kept separate so that PR stays reviewable.

**The ruff scope gap.** CI checked ruff against five explicit paths under `tools/`, so anything outside that list was never checked and drifted. Four files were unformatted: `check_agent_name_collisions.py`, both installers, and `yt-design-extractor.py`. Ruff now runs over all of `tools/`, those four are formatted, and two real `B007` findings in `yt-design-extractor.py` are fixed, where a loop unpacked a variable it never read.

`ty` keeps explicit paths rather than widening, because `yt-design-extractor.py` imports optional OCR dependencies that only `make install-ocr` provides, so `ty` cannot resolve them in CI. The paths now name every module it does cover, and there's a comment saying why the tool is excluded.

**A real type bug in `install_opencode.py`.** `default_config_dir` assigned `os.environ`, a `Mapping`, back over a parameter annotated `dict[str, str] | None`. That produced five `ty` errors and left the value reading as possibly-None for the next four lines. It now binds a separate `Mapping`-typed local. The other two installers were already fine.

**New `make lint` and `make format`.** Both run from `plugins/plugin-eval/`, which is where the ruff and ty config lives and where CI runs them. Running ruff from the repo root silently falls back to line-length 88 and disagrees with CI, which is easy to hit and produces reformatting CI does not want. The targets remove the choice.

I tried a root `ruff.toml` first and reverted it. It changes isort's first-party resolution for `tools/`, and CI's own command then fails with `I001` on two test files. Verified both ways before backing it out.

**Two stale docs.** AGENTS.md said generated artifacts are committed. Only the small registries are; the transformed trees under `.codex/`, `.opencode/`, `.copilot/` and `.antigravity/` are gitignored, which `.gitignore` states correctly. CONTRIBUTING listed `make garden STRICT=1` as a gate to run before pushing, but main carries ten `SKILL_OVER_CODEX_CAP` warnings and `STRICT=1` fails on warnings, so it has been red for a while. Two recent PRs noted this in their test plans. The gate list now shows plain `make garden`, with a note explaining what `STRICT=1` adds.

## Scope

- [x] CI / build / release
- [ ] Other

## Affected harnesses

- [ ] Claude Code
- [ ] OpenAI Codex CLI
- [ ] Cursor
- [ ] OpenCode
- [ ] Antigravity CLI

## Test plan

- [x] `make lint` — ruff check, ruff format --check and ty all clean over the new scope
- [x] `make validate STRICT=1` — OK across 5 harnesses
- [x] `make garden` — exits 0, unchanged output
- [x] `pytest tools/tests/` — 371 passed
- [x] markdownlint clean on CONTRIBUTING.md and AGENTS.md
- [x] `code-quality.yml` parses as YAML
- [x] AGENTS.md still 74 lines, under the 150 budget

The 3 `test_round_trip` failures are pre-existing and unrelated: `.opencode/` and `.copilot/` are gitignored and go stale locally. Identical on a clean checkout of main. CI regenerates them.

## Not addressed here

The ten oversize skills behind the `garden STRICT=1` warnings need their bodies split into `references/`. That is content work across ten plugins and belongs in its own PR. This change documents the situation accurately rather than hiding it.

## Related

Follow-up to #676. No overlap in files except `CONTRIBUTING.md`, which touches a different section.

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -35,34 +35,33 @@ jobs:
         run: uv sync --all-extras
 
       - name: ruff check (lint)
-        # Scope: adapter framework + plugin-eval. yt-design-extractor.py is legacy
-        # and out of scope for the multi-harness work; not gated.
+        # Scope: all of tools/ plus plugin-eval. Narrower per-file paths let
+        # check_agent_name_collisions.py, the installers and yt-design-extractor.py
+        # drift out of format; `make lint` runs this same scope locally.
         run: |
           uv run ruff check \
-            ../../tools/adapters/ \
-            ../../tools/generate.py \
-            ../../tools/validate_generated.py \
-            ../../tools/doc_gardener.py \
-            ../../tools/tests/ \
+            ../../tools/ \
             src/plugin_eval/
 
       - name: ruff format --check
         run: |
           uv run ruff format --check \
-            ../../tools/adapters/ \
-            ../../tools/generate.py \
-            ../../tools/validate_generated.py \
-            ../../tools/doc_gardener.py \
-            ../../tools/tests/ \
+            ../../tools/ \
             src/plugin_eval/
 
       - name: ty type-check
         run: |
+          # Not ../../tools/ — yt-design-extractor imports optional OCR deps that are
+          # installed only by `make install-ocr`, so ty cannot resolve them here.
           uv run ty check \
             ../../tools/adapters/ \
             ../../tools/generate.py \
             ../../tools/validate_generated.py \
             ../../tools/doc_gardener.py \
+            ../../tools/install_opencode.py \
+            ../../tools/install_copilot.py \
+            ../../tools/install_antigravity.py \
+            ../../tools/check_agent_name_collisions.py \
             ../../tools/tests/ \
             src/plugin_eval/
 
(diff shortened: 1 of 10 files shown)


wshobson:
Good catch, fixed in `afe5e9f`. The consequence is a bit worse than described, which makes it more worth fixing.

`ruff` and `ty` are declared under the `dev` optional dependency and my targets called `uv run` without it, so I expected a clean checkout to fail with "command not found". It doesn't. `uv run` provisions the tool on the fly, unpinned:

```
uv run --isolated --no-dev  ruff --version  ->  ruff 0.16.4
uv run --isolated --extra dev ruff --version ->  ruff 0.16.3
```

So `make lint` was silently running a newer ruff than CI's locked 0.16.3 rather than failing outright. For a target whose entire purpose is matching CI, quietly disagreeing about formatting is the worse failure of the two.

`--extra dev` is now on all five invocations, and the version resolves to 0.16.3. `make lint` passes: ruff check, ruff format --check and ty all clean over the widened scope.

I left `UV_TOOLS` alone. It runs `python` against `tools/`, which needs only the base dependencies.

wshobson:
@coderabbitai full review

wshobson:
@coderabbitai full review

PR #678, opened 2026-08-23
feat(plugin): add HOL Guard external marketplace entry

kantorcodes:
## Summary

- Add `hol-guard` as an external `git-subdir` marketplace plugin from `hashgraph-online/hol-guard-plugin/distributions/wshobson-agents`, resolving #654.
- Keep the approved local-only default explicit: Guard Cloud is neither required nor promoted, installation is gated on user approval, and the runtime hook/settings surface is disclosed.
- Update marketplace/catalog counts from 92 to 93 (91 local + 2 external) and document the external HOL Guard integration.

## Scope

- [ ] Plugin authoring (new or modified `plugins/<name>/...`)
- [ ] Adapter framework (`tools/adapters/`)
- [ ] Quality tooling (validators, gardener, plugin-eval)
- [x] Per-harness setup or docs (`docs/harnesses.md`)
- [x] AGENTS.md / ARCHITECTURE.md / docs/
- [ ] CI / build / release
- [x] Other: external marketplace `git-subdir` catalog entry

## Affected harnesses

- [x] Claude Code
- [ ] OpenAI Codex CLI
- [ ] Cursor
- [ ] OpenCode
- [ ] Antigravity CLI
- [ ] Pure tooling / framework (no harness behavior change)

The external source is a Claude marketplace `git-subdir` entry. `make generate-all` derives native registries from local `plugins/`, so no generated registry delta is expected; CI's generation/drift gates can verify that.

**Scope clarification:** this PR intentionally narrows #654's cross-harness discovery objective to Claude Code. It does not claim marketplace discovery or generated-registry inclusion for Codex, Cursor, OpenCode, Antigravity, Gemini CLI, or Copilot. The portable external skills may be usable elsewhere, but those harnesses are outside this contribution's accepted artifact.

## External/vendor disclosure

I maintain HOL Guard and `hashgraph-online/hol-guard-plugin`.

Submission-time payload reviewed by this PR:

- source: `https://github.com/hashgraph-online/hol-guard-plugin.git`
- path: `distributions/wshobson-agents`
- reviewed source commit: `43b2dda59e9f07057c52e69fd7426188faae1488`
- payload: `.claude-plugin/plugin.json`, `skills/hol-guard/SKILL.md`, `skills/plugin-scanner/SKILL.md`, `README.md`, `LICENSE`
- no `hooks/`, `.mcp.json`, install scripts, package lifecycle hooks, daemon, telemetry setup, OAuth flow, or hosted-service endpoint in the marketplace payload
[...]

Issue #654, opened 2026-08-14
[NEW] hol-guard — local agent runtime and supply-chain security


kantorcodes:
## Preliminary Checks

- [x] I have read the Code of Conduct.
- [x] I have read `docs/authoring.md` and the marketplace contribution workflow.
- [x] I reviewed `docs/plugins.md` and searched existing issues/PRs for HOL Guard / `hol-guard` / `plugin-scanner`.
- [x] This proposal is for legitimate, constructive use cases only.
- [x] I maintain the proposed open-source security project and its agent integrations.

## Component type

New plugin (external/source-referenced if maintainers prefer, following the existing `git-subdir` precedent)

## Proposed name

`hol-guard`

## Domain & expertise

Local security controls for autonomous coding agents: pre-tool/runtime policy, prompt-injection and exfiltration detection, unsafe command/package checks, and pre-install scanning of Agent Skills, plugins, MCP servers, and agent-tooling repositories.

Source/runtime: https://github.com/hashgraph-online/hol-guard
Portable plugin/skills source: https://github.com/hashgraph-online/hol-guard-plugin

HOL Guard is Apache-2.0 and the security decision can run locally; Guard Cloud is not required. I am a maintainer of the project.

## Unique value proposition

The marketplace already has strong components such as `security-scanning`, `block-no-verify`, and `protect-mcp`. This proposal is deliberately narrower in distribution shape and broader at the agent execution boundary:

- `security-scanning` focuses on SAST/vulnerability work in the codebase;
- `block-no-verify` blocks a specific class of hook-bypass commands;
- `protect-mcp` governs MCP tool calls with Cedar/receipts;
- HOL Guard provides a local pre-execution boundary across agent tool calls plus pre-install/supply-chain scanning for skills, plugins, MCP servers, packages, prompt-injection/exfiltration patterns, and unsafe commands.

The goal is not to duplicate those components. It is to expose an existing open-source runtime and its portable `SKILL.md` security workflows through this marketplace, ideally as an externally maintained/source-referenced plugin so implementation stays upstream.

## Primary use cases

1. **Before installing a skill/plugin/MCP server:** run the existing `plugin-scanner` skill locally against the repository and surface structured findings before trust/install.
2. **While an autonomous coding agent is running:** install/initialize HOL Guard and enforce local policy before risky tool/command/package actions.
3. **Agent supply-chain triage:** inspect prompt-injection, secret/exfiltration, unsafe-command, package, and provenance signals without requiring a cloud account.
[...]

PR #681, opened 2026-08-24
docs: use HOL Guard canonical product URL

kantorcodes:
## Summary

- Point the merged `hol-guard` marketplace entry's `homepage` at the canonical product page, `https://hol.org/guard`.
- Point the rendered README's HOL Guard product link at the same canonical page.
- Keep the external git-subdir source, reviewed SHA, version pins, and behavior unchanged.

This is a follow-up metadata/documentation correction to #678 and the accepted scope in #654.

## Validation

- `.claude-plugin/marketplace.json` parses as valid JSON after the change.
- No plugin payload, source URL, pin, install behavior, or generated harness registry is changed.

## Disclosure

I am President of HOL / Hashgraph Online and am materially affiliated with HOL Guard.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -1130,7 +1130,7 @@
         "name": "Hashgraph Online",
         "url": "https://github.com/hashgraph-online"
       },
-      "homepage": "https://github.com/hashgraph-online/hol-guard",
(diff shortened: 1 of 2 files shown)


wshobson:
Merged, thanks for the quick follow-up.

PR #691, opened 2026-08-31
Add document-trust plugin: forensic document verification agent

Sketchjar:
""" + body + """

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -1279,6 +1279,11 @@
         "deck-generation",
         "slide-design"
       ]
+    },
+    {
+      "name": "document-trust",
+      "source": "./plugins/document-trust",
+      "description": "Forensic document authenticity verification, citation checking, grounded extraction, AI-text detection, identity checks, adverse-media screening, and tender search"
     }
   ]
 }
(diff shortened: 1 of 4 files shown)


wshobson:
Closing this one.

The plugin routes every check through Stipple, and Stipple is a metered service: a 7-day, 200-credit trial, then credit packs from $5, with a small anonymous weekly allowance in front of it. That is the shape the "External and vendor plugins" section of CONTRIBUTING rules out on the default path, and the reason #648 was closed. It matters more here because the documents being uploaded are passports, IDs, and bank statements.

Your GitHub profile links stipple.sh, so this is a plugin for a service you operate, and neither the PR nor the plugin discloses that. The PR body came through as a template placeholder, and the same submission went to roughly thirty other repositories on the same day, which puts it under the promotion clause of the commercial section.

For the record, it would also fail CI as-is: `make garden` reports 10 `STALE_COUNT` errors because the README and AGENTS counts were not bumped, `make generate-all` produces registry drift, the marketplace entry is missing `version`, `author`, `license`, and `category`, there is no plugin README, and the description promises tender search and an MCP screening tool that nothing in the payload provides.

The policy bar is in CONTRIBUTING under "Commercial content and disclosure" and "External and vendor plugins".

PR #693, opened 2026-09-01
feat: support gh skill and npx skills installers

wshobson:
## Summary

- Adds first-class support for the two Agent Skills installers, `gh skill install wshobson/agents` (GitHub CLI 2.90+) and `npx skills add wshobson/agents` (vercel-labs/skills). Both already discover all 182 skills through the `plugins/<plugin>/skills/<skill>/` layout, so this is docs plus a real-CLI gate, not a layout change: README quick start, a `Skills-only installers` section in docs/harnesses.md (selectors, install paths, the release-freeze rule, the local-checkout caveat), a pointer in docs/agent-skills.md, one AGENTS.md bullet, and an authoring note that a skill's `name` must equal its directory name.
- Four new smoke tests in `tools/tests/test_cli_smoke.py`: gh local discovery lists every source skill, `gh skill publish --dry-run` passes the agentskills.io spec check, npx discovery lists every source skill, and skill directory names stay unique across plugins (both installers install under the bare skill name). The smoke job sets up Node for npx and fails loudly if the runner's gh predates `gh skill` instead of skipping silently.
- Renames `plugins/database-design/skills/postgresql` to `postgresql-table-design`. Its frontmatter already declared that name, and the spec requires the directory to match; it was the one error `gh skill publish --dry-run` reported. Claude Code, `gh skill`, and `npx skills` all key on the frontmatter name, so the skill's identity in those channels is unchanged; the gitignored per-harness trees derive IDs from the directory name, so `database-design__postgresql` becomes `database-design__postgresql-table-design` on the next `make generate-all`. No committed registry drift.
- Follow-up commit, droppable: three PostgreSQL facts in that skill were wrong (`NULLS NOT DISTINCT` placement, a non-built-in `current_user_id()`, partitioned-table foreign keys). Review flagged them once the rename made the file look new; fixed in a separate commit.

- Follow-ups from review: npx parser tightened to the exact list line shape; CI upgrades gh from the official apt repository when the runner build predates `gh skill`; every doc that covers installation now names the installers, and docs/authoring.md records that a skill directory name is an identity for installers and adapters.

## Scope

- [x] Plugin authoring (new or modified `plugins/<name>/...`)
- [x] Per-harness setup or docs (docs/harnesses.md)
- [x] CI / build / release
- [ ] Other

## Affected harnesses

- [x] Claude Code
- [x] OpenAI Codex CLI
- [x] Cursor
- [x] OpenCode
- [x] Antigravity CLI
- [ ] Pure tooling / framework (no harness behavior change)

Skills-only distribution reaches every agent both CLIs support; nothing in the generated per-harness trees changes.

## Test plan

- [x] `make validate STRICT=1` (OK across 5 harnesses)
- [x] `make garden` (0 errors, the 21 pre-existing warnings)
- [x] `make test` (601 passed)
- [x] `make smoke-test` (12 of 12; the OpenCode agent-list flake is root-caused and fixed here: the CLI exits before a pipe drains its ~330k-line output, so both OpenCode tests now capture through a file)
- [x] Ran ruff + ty in `plugins/plugin-eval/` (`make lint` clean)
- [x] Spot-checked behavior in a real harness (which?): `gh skill install wshobson/agents python-testing-patterns --dir <tmp>` and the exact-path form both install; the namespaced `python-development/python-testing-patterns` selector does not resolve, which is why the docs use the bare name. `npx skills add wshobson/agents --skill python-testing-patterns -a claude-code` installs to `.claude/skills/`. `gh skill publish --dry-run` exits 0 unauthenticated after the rename.

- The skill is now split under the Codex 8 KB cap: SKILL.md keeps When to Use, the rules, and quick-start DDL; references/details.md carries the rest verbatim. No committed registry changes.
- gh resolves the install ref as latest GitHub release first, `main` second. The repo has no releases, so installs track `main`; creating a release would freeze `gh skill` installs at that tag (documented).

## Related issue(s)

N/A

https://claude.ai/code/session_01LjJmzuuxXSwGNEYdBvsmFs


wshobson:
Review responses, all pushed.

**Codex P1, generated skill IDs.** Correct that the gitignored harness trees key on the directory name: `database-design__postgresql` becomes `database-design__postgresql-table-design` on the next `make generate-all`, and the pruning is by design. The spec requires the directory and the frontmatter name to agree, so one of them had to move. The frontmatter name is what Claude Code, `gh skill`, and `npx skills` all install and invoke by, and those are the channels this PR is about, so the directory is the one that moved. The generated trees are rebuilt locally and never committed, and the adapters have no alias mechanism to preserve a stale ID; adding one for a single skill is not worth it. The PR body now says this instead of "identity does not change".

**CodeRabbit, three content findings in the skill.** This PR only moves the file, so the review saw it as new. The findings are real: the `NULLS NOT DISTINCT` clause belongs before the column list, `current_user_id()` is not a built-in, and foreign keys on partitioned tables have worked since PG11 (from) and PG12 (referencing). Fixed in aee6254 as a separate commit so it can be dropped if the rename should stay content-free. The `env` annotation nit is in 5f778a1.

**Claude review.** The 8 KB claim in the PR body was wrong; the skill body is about 16 KB and stays on the pre-existing over-cap list. Corrected in the body. The npx regex is a containment check and passed in CI; leaving it unless it flakes.

wshobson:
@coderabbitai review

wshobson:
Second round, pushed as a1eff22, daa8887, 1e87200.

- **npx parser** now matches only the `│    <name>` line shape, so a stray lowercase token cannot count as a skill (Claude nit 1).
- **gh runner dependency** removed: the smoke job upgrades GitHub CLI from the official apt repository when the preinstalled build predates `gh skill`, and the version step still hard-fails if that somehow leaves it missing (Claude nit 2).
- **postgresql-table-design over the Codex cap** (Claude nit 3, pre-existing): split into a 7.7 KB SKILL.md carrying When to Use, the rules and decision points, and three quick-start DDL examples, with references/details.md holding the data-type catalog, workload patterns, extensions, JSONB indexing, and partitioning DDL verbatim. Clears the skill's SKILL_OVER_CODEX_CAP warning, so garden is at 20 warnings from 21. plugin-eval moves from 0.60 to 0.65 on progressive disclosure and 0.55 to 0.70 on structural completeness.
- **OpenCode discovery flake** (pre-existing): `opencode agent list` prints every agent's expanded permission array, about 330k lines, and exits before a pipe is drained, so pipe capture lost the alphabetically last agents in 1 of 5 measured runs while a file sink lost none in 5 of 5. Both OpenCode smoke tests now capture stdout through a file.
- **Docs sweep**: every place that documents installation now names the installers: the README multi-harness section, docs/usage.md, docs/plugins.md, docs/architecture.md, ARCHITECTURE.md, the docs/harnesses.md supported table, the CONTRIBUTING portability checklist, and docs/round-trip-results.md summary rows plus a reproduce recipe. docs/authoring.md states that a skill directory name is an identity for installers and adapters alike, which is the Codex P1 point made durable.

Gates at 1e87200: `make validate STRICT=1` clean, `make garden` 0 errors, `make test` 601 passed, `make lint` clean, smoke 12 of 12 including the four installer tests and both OpenCode tests.

PR #694, opened 2026-09-01
fix: issue triage — grounded-vault skill, $ARGUMENTS framing, agent copy reconciliation

### LINKED ISSUE #673: Idea: grounded vault pattern (raw/wiki/archive + Fingerprint) as a skill?
Hi team — love the marketplace structure (181 skills, progressive disclosure is spot on).

We have been running a small 3-layer markdown vault in production:

- 
aw/ immutable (notes, papers, logs — never edited by agent)
- wiki/ compiled with > Raw: links for every number/quote
- rchive/ for Status: Outdated after drift

Plus a tiny Fingerprint: git:<hash> + Monitored: header so git diff checks drift in 0.01s instead of re-reading the codebase.

We wrapped it as a local skill (wiki-manager) and it cut our re-read tokens a lot. Would a grounded-vault skill be useful here? Happy to draft it under plugins/knowledge/ following your gent_skills_spec if you think it fits. No pressure — just wanted to ask before opening a PR.

Repo & live demo for context (MIT, zero-dep, 25 tests): https://github.com/PALAN-K/llm-wiki-loop

### LINKED ISSUE #688: Commands interpolate raw $ARGUMENTS into tool-enabled agent prompts
Split out from #643, where @sergiopadure raised it as a side note.

Some command templates place the raw `$ARGUMENTS` string directly into a prompt that runs with tool access, e.g., `plugins/dgx-spark-ops/commands/spark-preflight.md`. If the argument text comes from an untrusted source, the text can carry instructions that the agent then acts on with tools.

Two things to decide.

- Whether the authoring guide or CONTRIBUTING should say how to handle `$ARGUMENTS` in commands that run with tools, e.g., treat it as data and quote it.
- Whether garden or plugin-eval should flag commands that interpolate `$ARGUMENTS` into a tool-enabled prompt without any framing.

wshobson:
## Summary

- **Closes #673.** Adds the `grounded-vault` skill to `documentation-standards`, shaped the way the issue thread asked: it teaches the raw/wiki/archive layout, per-claim `> Raw:` links, the `Fingerprint`/`Monitored` page header with a one-`git diff` drift check, the archive rule, and a commit gate, with a standard-library check script and templates in `references/`. No dependency on `llm-wiki-loop`; it is linked once as the reference implementation and the pattern is credited to the issue author. The plugin goes to 1.1.0 with a description that names both skills; catalog rows and every skill count move to 183; registries regenerated.
- **Closes #688.** Both questions answered: `docs/authoring.md` gains "Treat `$ARGUMENTS` as data" (block and inline shapes) and CONTRIBUTING's checklist points at it; `make garden` gains `ARGUMENTS_UNFRAMED` (warning, `--check arguments`), which flags a command that interpolates the token into prompt text with no `<user_request>` block, no nearby data-not-instructions sentence, and no backticked reference, skipping fenced code. All 79 commands that use `$ARGUMENTS` are framed in two commits (37 template commands plus onboard and the spark-preflight example, then the 30 inline uses in 16 commands), so the check reports zero today and guards new commands.
- **Advances #643.** The two divergent groups that were strict supersets (`django-pro`, `deployment-engineer`) are reconciled, taking `AGENT_BODY_DIVERGENT` from 11 to 9. The remaining nine split into three tier-only differences and six intentional variants; both kinds need a maintainer decision and are laid out on the issue. The `#676` rule that a `model:` difference is drift is left as is.

## Scope

- [x] Plugin authoring (new or modified `plugins/<name>/...`)
- [ ] CI / build / release
- [ ] Other

## Affected harnesses

- [x] Claude Code
- [x] OpenAI Codex CLI
- [x] Cursor
- [x] OpenCode
- [x] Antigravity CLI
- [ ] Pure tooling / framework (no harness behavior change)

Command bodies and one new skill ship to every harness; the framing text is plain Markdown with no tool vocabulary.

## Test plan

- [x] `make validate STRICT=1` (OK across 5 harnesses)
- [x] `make garden` (0 errors, 19 warnings, down from 21; `ARGUMENTS_UNFRAMED` 0)
- [x] `make test` (609 passed; 9 new gardener tests, written first and watched fail)
- [ ] `make smoke-test` (the installer smoke tests live in #693; on this branch `gh skill install . --from-local` lists 183 skills and `gh skill publish --dry-run` reports no error for the new skill)
- [x] Ran ruff + ty in `plugins/plugin-eval/` (`make lint` clean)
- [x] Spot-checked behavior in a real harness (which?): the vault check script from the skill's `references/details.md` was extracted and run against a fixture vault; it caught the one ungrounded figure and exited 1 under `--strict`.

- `grounded-vault` SKILL.md is a 5 KB body with `references/details.md`, under the Codex cap; plugin-eval quick score 0.71 triggering, 0.65 progressive disclosure.
- Renamed nothing, so no generated IDs change.

## Related issue(s)

Closes #673
Closes #688
Refs #643

https://claude.ai/code/session_01LjJmzuuxXSwGNEYdBvsmFs


wshobson:
Review responses, pushed as 3f8710a.

**Vault check script (CodeRabbit items 2 to 7, Claude's inline).** All real, all fixed in `references/details.md` and proven on a fixture vault: link labels and paths, headings, the header block, and fenced code are excluded from claim scanning, so `raw/adr/0007-jwt.md` no longer reads as a claim of `0007`; numbers match as whole tokens, so `15` is not `150` or `2015`; a linked source must resolve inside `raw/`, so a traversal path or a missing file is a miss; under `--strict` a number or quotation with no `raw/` link is an error; a page without `Fingerprint:` is an error; and a git failure (unknown fingerprint after a history rewrite) counts as drift instead of being swallowed. Two parts kept as they were, and documented: quoted phrases shorter than eight characters are not treated as claims, because a two-word quote is not evidence, and an empty `Monitored:` stays legal because a page compiled only from `raw/` has no code to drift against.

**"Remove framing-only security guidance" (CodeRabbit item 1).** Not taken as written. Claude Code substitutes `` into the prompt textually; there is no trusted argument channel for a command to use, and confirmation before tool actions is the harness's permission model, not something a command's Markdown can enforce. Framing is the mitigation #688 asked for and the only one available at the command layer. What the guidance must not do is overclaim, so `docs/authoring.md` now says plainly that framing lowers the chance the model follows injected text, is not a security boundary, and that tool permissions and approval prompts remain the control. `tools/adapters/antigravity.py` is untouched; it transpiles command bodies and has no argument channel either.

**Claude's nit on coarse number matching.** Correct and now narrower with whole-token matching; still coarse by design (a figure that exists anywhere in the source grounds), which the reference file states.

@coderabbitai review

PR #700, opened 2026-09-07
fix(adapters): quote YAML scalars in OpenCode and Copilot frontmatter

MaxFreedomPollard:
## Summary

- `_opencode_frontmatter` (`tools/adapters/opencode.py:132`) writes every scalar bare, and `_copilot_frontmatter` (`tools/adapters/copilot.py:32`) quotes only pure numbers and lowercase reserved words through `_needs_yaml_quoting`. Source frontmatter that is already correctly quoted loses its quotes on the way out, so both harnesses get files a YAML parser rejects.
- Codex and Antigravity never had this problem: both already route scalars through a private `_yaml_scalar` helper, duplicated verbatim between `codex.py` and `antigravity.py`. This moves that helper into `tools/adapters/base.py` as `yaml_scalar` and has all four adapters use it.
- Copilot also had no `dict` branch, so a mapping-valued field like `metadata:` came out as a Python dict repr. It now emits a nested mapping, matching `_antigravity_frontmatter`.

## The bug

`plugins/conductor/commands/manage.md` starts with valid, correctly quoted YAML:

```yaml
description: "Manage track lifecycle: archive, restore, delete, rename, and cleanup"
argument-hint: "[--archive | --restore | --delete | --rename | --list | --cleanup]"
```

`make generate HARNESS=opencode` turns it into this, which `yaml.safe_load` refuses with `ScannerError: mapping values are not allowed here`:

```yaml
description: Manage track lifecycle: archive, restore, delete, rename, and cleanup
argument-hint: [--archive | --restore | --delete | --rename | --list | --cleanup]
```

The second failure mode is silent. When a bracketed hint has no colon inside it, YAML accepts it as a flow sequence, so `argument-hint: "[--resume]"` round-trips to the list `['--resume']` instead of the string `[--resume]`.

I generated every plugin for all four harnesses from `a30778f` and ran `yaml.safe_load` over each emitted frontmatter block:

| harness | blocks | unparseable | mistyped |
|---|---|---|---|
| opencode | 490 | 8 | 12 |
| copilot | 687 | 30 | 24 |
| codex | 288 | 0 | 0 |
| antigravity | 385 | 0 | 0 |

The repo's own gates do not catch it because `validate_generated.py` and `doc_gardener.py` read frontmatter with `parse_frontmatter`, the tolerant hand-rolled reader in `base.py`, not with a YAML parser.

## The fix

`yaml_scalar` quotes a value when it is empty or pure whitespace, starts with a YAML special character, contains `: ` or ` #`, has leading or trailing whitespace, starts with a digit, is an implicit number led by `+` or `.` such as `+1`, `.5`, `.inf` or `.nan`, or is a YAML 1.1 implicit boolean or null. Every clause except the implicit-number one is the existing Codex and Antigravity behavior, moved rather than rewritten; no value in the repo today hits the implicit-number clause, so those two adapters are unaffected and verified byte identical below.

## Test plan

- [x] `make validate STRICT=1` gives `OK: no issues across 5 harness(es).`
- [x] `make garden` gives `0 error(s), 18 warning(s)`, identical to `a30778f` with the change stashed (9 `SKILL_OVER_CODEX_CAP`, 9 `AGENT_BODY_DIVERGENT`, all pre-existing).
- [x] `make test` gives `619 passed`.
- [x] `make lint` gives ruff check `All checks passed!`, ruff format `41 files already formatted`, ty `All checks passed!` (with `uv sync --all-extras`, as the code-quality workflow does).
- [x] `make generate-all` leaves the committed registries unchanged, so there is no drift.
- [ ] `make smoke-test`: 6 passed, 5 skipped (opencode, agy and codex are not on this machine), 1 failed in `test_gh_skill_discovers_every_source_skill` because my `gh` does not accept `gh skill install . --from-local`. That failure reproduces identically on `a30778f` with the change stashed, so it is my GitHub CLI version, not this PR.

Byte-identical output check for the two adapters I only refactored, generating every plugin before and after into separate trees:

```
diff -r base-codex fix-codex            # no differences, 908 files
diff -r base-antigravity fix-antigravity  # no differences, 824 files
```

After the change all four harnesses report 0 unparseable and 0 mistyped. The only output differences are 37 opencode files (38 lines) and 77 copilot files (81 removed, 87 added), and every one is a frontmatter value gaining quotes, a `version:` like `1.0.0` becoming `"1.0.0"`, or Copilot's `metadata:` dict repr expanding into a nested mapping.

`tools/tests/test_adapters.py` gains a `TestFrontmatterYamlSafety` class that parses emitted frontmatter with PyYAML instead of `parse_frontmatter`, using the two real shapes above. Running that same emission through the unmodified `a30778f` adapters, all four generated files raise `ScannerError`: copilot's SKILL.md, index.md and manage.md, and opencode's `demo__manage.md`. I replaced `TestCopilotAdapter.test_yaml_quoting`, which asserted on the now-deleted `_needs_yaml_quoting`, with a `yaml_scalar` case covering everything it did plus the shapes it missed.

## Scope

- [x] Adapter framework (`tools/adapters/`)
- [ ] Quality tooling (validators, gardener, plugin-eval)
- [ ] CI / build / release
- [ ] Other

## Affected harnesses

- [ ] Claude Code
- [ ] OpenAI Codex CLI
- [ ] Cursor
- [x] OpenCode
- [ ] Antigravity CLI
- [x] GitHub Copilot

Codex and Antigravity are touched but their output is unchanged.

N/A, no plugin content changed.

## Related issue(s)

None. This is a self-contained adapter fix with a regression test.

### DIFF
--- a/tools/adapters/antigravity.py
+++ b/tools/adapters/antigravity.py
@@ -37,6 +37,7 @@ from tools.adapters.base import (
     HarnessAdapter,
     PluginSource,
     SkillSource,
+    yaml_scalar,
 )
 from tools.adapters.capabilities import TOOL_NAME_MAPS, resolve_model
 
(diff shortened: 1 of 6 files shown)


wshobson:
Thanks Max, this is a clean fix and the write-up made it easy to verify. I reproduced the 9 OpenCode and 31 Copilot unparseable blocks on main and got 0 with your branch. The second commit covers all three CodeRabbit findings, CI is green after workflow approval, and I filed #704 so the gardener parses generated frontmatter with a real YAML loader going forward. Merging shortly.

Issue #705, opened 2026-09-13
protect-mcp and review-agent-governance: Cedar policies use an entity shape that protect-mcp 0.7 does not evaluate


wshobson:
protect-mcp 0.7.x evaluates every tool call as `action == Action::"MCP::Tool::call"`, `resource == Tool::"<tool name>"`, with the tool input at `context.input` (for Bash, `context.input.command`). The test policy at plugins/protect-mcp/test/fixtures/test-policy.cedar was rewritten to that shape in #602 and its comment explains the change.

Every other policy and policy example in both plugins still uses the older shape: `action == Action::"Bash"`, `Action::"Read"`, and so on, with `context.command_pattern`. Against 0.7.4 those rules never match, so the trailing `permit (principal, action, resource)` allows every call. I confirmed it directly: `evaluate --policy plugins/review-agent-governance/policies/review-agent-governance.cedar --tool Bash --input '{"command":"gh pr merge"}'` returns allowed, exit 0.

Files with the old shape:
- plugins/review-agent-governance/policies/review-agent-governance.cedar and .cedarschema
- plugins/review-agent-governance/agents/review-policy-author.md and README.md
- plugins/protect-mcp/skills/protect-mcp-setup/SKILL.md, agents/policy-enforcer.md, and README.md

Found while fixing #701. That PR fixes the hook wiring and keeps the policies unchanged, so its governance test covers the allow path and the approval flag only.

Suggested fix: rewrite the shipped policy, the schema, and every example to the 0.7 shape, match Bash commands with `context.input.command like "gh pr merge*"` so commands with arguments are caught, and add a test in plugins/review-agent-governance/test/run-tests.sh that pipes a `gh pr merge 42` payload through the hook and expects exit 2. The README's list of gated commands should be checked against the policy at the same time.

PR #706, opened 2026-09-13
fix(hooks): read the Claude Code hook payload from stdin in protect-mcp and review-agent-governance

### LINKED ISSUE #701: [BUG] protect-mcp / review-agent-governance hooks pass $TOOL_NAME and $TOOL_INPUT, which Claude Code does not set
### Preliminary Checks

- [x] I have read the Code of Conduct
- [x] I have searched existing issues to ensure this is not a duplicate. The closest prior art is #601, same plugin family but a different root cause: that one was about subcommands missing on `@0.5.5`, and they exist on `@0.7.4`.

### Harness

Claude Code

### Affected plugin / agent / skill / command

`plugins/protect-mcp/hooks/hooks.json` and `plugins/review-agent-governance/hooks/hooks.json`, both the `PreToolUse` and `PostToolUse` entries.

### Bug description

Both hook files pass the tool name and tool input to `protect-mcp` through the shell variables `$TOOL_NAME` and `$TOOL_INPUT`. Claude Code never sets those variables.

The hooks documentation says a `command` hook gets its payload as JSON on stdin, and that the only environment variables Claude Code sets on the hook process are `CLAUDE_PROJECT_DIR`, `CLAUDE_PLUGIN_ROOT`, `CLAUDE_PLUGIN_DATA` and `CLAUDE_EFFORT`. `TOOL_NAME` and `TOOL_INPUT` are not on that list, so `--tool "$TOOL_NAME"` expands to an empty string and `--input "$TOOL_INPUT"` falls back to `"{}"`.

`protect-mcp` does read stdin, but only behind `--format`, and neither hook file passes it. From `dist/cli.mjs` in `protect-mcp@0.7.4`:

```js
3314:  let tool = flagValue(argv, "--tool") || "";
3315:  let inputRaw = flagValue(argv, "--input") || "{}";
3317:  const failOnMissing = flagValue(argv, "--fail-on-missing-policy") !== "false";
3318:  if (format) {                       // stdin is only read when --format is set
3319:    const j = await readHookStdin();
```

So the policy always gets an empty tool name and empty input, whatever the real call was.

That produces two failure modes, and neither of them is visible to the user:

1. With no policy file present, which is the default since there is no `./protect.cedar` or `./review-governance.cedar`, `--fail-on-missing-policy false` makes the hook exit 0 and allow everything. The plugin is installed and enforces nothing.
2. With a policy file present, the tool name is empty, no `permit` rule matches, and Cedar denies by default. The hook exits 2, which is how Claude Code blocks a call, so every tool call gets blocked including the ones the policy explicitly permits.

### Steps to reproduce

Measured against `protect-mcp@0.7.4` with its optional dependencies installed (`@cedar-policy/cedar-wasm@4.12.0`) on Node v22.23.2, using this repo's own fixtures:

```bash
POL=plugins/protect-mcp/test/fixtures/test-policy.cedar
CLI=node_modules/protect-mcp/dist/cli.mjs   # or npx protect-mcp@0.7.4

# A) Correct tool name, permitted by the policy
node $CLI evaluate --policy $POL --tool Read --input '{}' --fail-on-missing-policy false; echo "exit=$?"

# B) Correct tool name, forbidden by the policy
node $CLI evaluate --policy $POL --tool Write --input '{"file_path":"./x"}' --fail-on-missing-policy false; echo "exit=$?"

# C) Exactly as hooks.json invokes it, with TOOL_NAME/TOOL_INPUT unset
node $CLI evaluate --policy $POL --tool "$TOOL_NAME" --input "$TOOL_INPUT" --fail-on-missing-policy false; echo "exit=$?"

# D) Same call, but with --format and the real hook payload on stdin
cat plugins/protect-mcp/test/fixtures/pretool-allow-read.json \
  | node $CLI evaluate --policy $POL --format claude --fail-on-missing-policy false; echo "exit=$?"
```

### Expected behavior

C should behave like A. The hook payload for a `Read` call should reach the policy, match the `permit` rule for read-oriented tools, and exit 0.

### Actual behavior

```
A) {"allowed":true,"policy_digest":"5c5b7ea861cc5d1a"}
   exit=0

B) {"allowed":false,"reason":"cedar_deny: {\"reason\":[\"policy3\"],\"errors\":[]}", ...}
   exit=2

C) {"allowed":false,"reason":"cedar_deny: {\"reason\":[],\"errors\":[]}", ...}
   exit=2

D) {"allowed":true,"reason":"allowed"}
   exit=0
```

C denies with an empty `reason` array. Nothing matched, so this is Cedar's implicit deny on an empty tool name rather than a policy decision, and exit 2 blocks the call. D is the same underlying call with `--format` added and the payload on stdin, and it gives the right answer, which points at the wiring rather than the policy or the evaluator.

### Additional context

The existing tests miss this because `plugins/protect-mcp/test/run-tests.sh` does not invoke the hook the way Claude Code does. It reads each fixture with Python and passes the fields as CLI arguments:

```bash
--tool "$(extract "$INPUT" tool_name)" \
--input "$(python3 -c '...json.dumps(...["tool_input"])' "$INPUT")"
```

That covers the CLI, which works fine. Nothing covers the bridge between Claude Code and the CLI. A test that pipes a fixture to stdin, using the hook command copied verbatim from `hooks.json`, would have caught it.

For the fix, dropping `--tool` and `--input` from both files and passing `--format` should be enough to engage the stdin path:

```
npx protect-mcp@0.7.4 evaluate --policy "${PROTECT_MCP_POLICY:-./protect.cedar}" --format claude --fail-on-missing-policy false
```

D above shows that path returning the right decision and the right exit code. The `PostToolUse` `sign` command needs the same treatment, since it reads stdin under the same `if (format)` guard at `dist/cli.mjs:3363`.

Two smaller things I noticed while digging, both lower priority than the above:

`--fail-on-missing-policy false` means a fresh install enforces nothing and says nothing about it. A one-line warning on stderr when the policy file is absent would at least make that state visible, which seems worth it for a plugin whose job is enforcement.

Blocking currently rests on exit code 2 alone. Emitting `hookSpecificOutput.permissionDecision: "deny"` with a `permissionDecisionReason` would pass the policy reason back to the model instead of only to stderr. Blocking works without it.

On how I verified this: I read the hooks documentation and the `protect-mcp@0.7.4` sources first, then ran the four commands above to check the behaviour rather than assume it. Those runs are standalone CLI invocations, not a live Claude Code session, so the exit codes are measured but the end-to-end behaviour inside a session is inferred from them plus the documented meaning of exit 2.

wshobson:
Summary: both plugins' hooks.json passed `--tool "$TOOL_NAME" --input "$TOOL_INPUT"` to protect-mcp. Claude Code sends the hook event as JSON on stdin and never sets those variables, so the policy always saw an empty tool name. With no policy file everything was allowed, and with a policy file everything was denied. Reported in #701.

Change: each plugin now ships `hooks/evaluate.sh` and `hooks/sign.sh`, and hooks.json runs them through `${CLAUDE_PLUGIN_ROOT}`. The scripts read `tool_name` and `tool_input` from stdin with node, which npx already requires, and pass them to protect-mcp 0.7.4 as flags. That is the only input path the 0.7.4 CLI documents for Claude Code, since no release has a `--format claude` (upstream ask: ScopeBlind/scopeblind-gateway#10). Skill docs updated to match. Plugin versions 0.1.1 to 0.1.2.

Behavior changes: a missing policy file now prints a warning to stderr and still allows the call. An unreadable policy file now denies, which is the CLI default, instead of allowing. `sign` no longer passes `--input` and `--output`, because 0.7.4's sign ignores them.

Tests: protect-mcp's run-tests.sh gains three tests that read the command string out of hooks.json and run it with a fixture on stdin, the way Claude Code does. 11 passed. review-agent-governance's run-tests.sh gains Part C with the same shape (allow Read through the policy, and the approval flag short-circuits before the policy is read). `make validate STRICT=1` OK, `make garden` 0 errors and 18 warnings (unchanged), `make generate-all` leaves only the version bumps in the registries.

Not in this PR: the shipped governance policy and every policy example in both plugins use an entity shape that protect-mcp 0.7 does not evaluate, so they permit everything even with correct wiring. Filed as #705. Fixing that needs a policy rewrite and its own review.

Closes #701.


wshobson:
Thanks CodeRabbit. Two of the four findings are fixed in the latest commit. Both evaluate.sh scripts now map an npx exit other than 0 or 2 to exit 2, so a huge tool input that stops the evaluator from running blocks the call instead of letting it through, and test 12 covers that with a fake npx. The setup skill now says sign.sh reads only tool_name.

The other two stay as they are. The governance policy's entity shape is a real problem, and it is tracked separately in #705 because it needs a policy rewrite of its own. The .cursor-plugin files were produced by `make generate-all` for the version bump, not edited by hand.

PR #707, opened 2026-09-13
feat: add the Pi harness (pi.dev) as a sixth generated target

wshobson:
Summary: this branch adds the Pi coding agent (pi.dev, npm package `@earendil-works/pi-coding-agent`, binary `pi`) as a sixth generated target, alongside Codex CLI, Cursor, OpenCode, Antigravity CLI, and Copilot. Claude Code remains the native source of truth. Nothing under `plugins/` changed, so every harness still comes from the same Markdown source.

What Pi reads. Pi implements the Agent Skills standard and discovers `SKILL.md` recursively, so a skill can sit at any depth under the skills directory. Slash commands are plain Markdown prompt templates, discovered flat, one file per command, with `$ARGUMENTS` substituted by Pi at expansion time. Pi core has no subagents, so agents are written in the format used by the reference `subagent` example extension that ships inside the pi package, which is name, description, tools, and model in the frontmatter with the body as the system prompt.

The generated tree is `.pi/` and it is gitignored like the other transformed trees. Skills land at `.pi/skills/<plugin>/<skill>/SKILL.md`, prompt templates at `.pi/prompts/<plugin>__<command>.md`, and agents at `.pi/agents/<plugin>__<agent>.md`. Prompts and agents carry the `<plugin>__<name>` prefix because Pi discovers both flat and the filename is the identifier, so two plugins with the same command stem would otherwise collide. Skills stay bare because the per plugin directory already keeps them apart, recursive discovery finds them anyway, and skill names are already unique across plugins. That also keeps the skill name the user types the same on Pi as it is everywhere else.

The agents caveat. Our generated agent files do nothing on a stock Pi install. Pi has no agents resource of its own, and the reference `subagent` extension or a compatible package has to be installed before these files mean anything. The docs say this in every place the Pi agents are mentioned, and the follow up list below records that we have not yet confirmed the extension dispatches them end to end.

What was added. There is a new adapter at `tools/adapters/pi.py`, a `make generate HARNESS=pi` path, Pi coverage in `tools/validate_generated.py`, a `.pi/` mapping in the gardener so stale artifacts are reported, a `make install-pi` target that symlinks every generated skill directory, prompt template, and agent into the Pi config dir, real CLI smoke tests in `tools/tests/test_cli_smoke.py`, a CI step that installs pi pinned to 0.85.1, and Pi rows in the README, `ARCHITECTURE.md`, `AGENTS.md`, `docs/harnesses.md`, `docs/authoring.md`, and `docs/round-trip-results.md`.

How the smoke test proves discovery without billing tokens. Pi expands `/template args` and `/skill:name` into the user message before it calls the model, so the expanded text in the json stream is the proof that Pi found the artifact. The test sets every Anthropic credential variable to an invalid value and also points `ANTHROPIC_BASE_URL` at a closed port, because a developer aimed at a gateway that authenticates on its own would otherwise pay for the call despite the bad keys. With the base URL closed the model call fails before any request is completed, so no tokens are billed, and the expansion we assert on has already happened by then.

Gate results from a clean run on this branch. `make generate-all` wrote 732 files for Pi with 0 skipped, 0 warnings, 0 errors, and 0 pruned, which covers 183 skills, 105 prompt templates, 202 agents, and the reference files that skills carry. `make validate STRICT=1` reported `OK: no issues across 6 harness(es).` `make garden` reported `Totals: 0 error(s), 18 warning(s), 0 info.` `make lint` passed ruff check, reported 44 files already formatted, and passed ty. `make test` passed 655 tests in 12.58 seconds. `make smoke-test` passed all 15 tests in 192.24 seconds with the three `TestPiSmoke` tests running rather than skipping. A second run of that class alone took 171.63 seconds and broke down as 61.92 seconds to expand all 105 prompt templates, 109.00 seconds to expand all 183 skills, and 0.63 seconds for the negative test that checks an unknown template stays literal. The working tree was clean afterwards.

The design spec lives at `docs/superpowers/specs/2026-09-13-pi-harness-design.md`. That path is gitignored, so the file is local only and is not part of this pull request.

Follow-ups. The shared prose tool rewrite still emits `open`, `rg`, and `glob` in skill and agent prose, while Pi's own tools are named `read`, `grep`, and `find`, so that mapping should be extended for Pi. `make install-pi` does not prune dangling symlinks when a source artifact goes away, which is the same gap Antigravity has today. The harness list in `.github/workflows/claude-code-review.yml` omits Copilot and its counts are stale, which is a pre existing issue that this branch did not introduce and did not fix.

https://claude.ai/code/session_01DZazzWVyb8MxPCuLC1w5Qo


wshobson:
Thanks for the review. I applied the findings in 820734d.

The main one was the cleanup scope. `.pi/` is also Pi's project-local config directory, so a developer may keep `.pi/settings.json` or `.pi/extensions/*.ts` there. Clean and prune now stay inside `.pi/skills`, `.pi/prompts` and `.pi/agents`, which are the only subtrees the adapter writes. There is a new regression test that plants a settings file and an extension next to a generated tree, runs both clean and prune, and checks that the planted files survive while a stale prompt is removed. The README output table, docs/harnesses.md and AGENTS.md now say the same thing.

The Pi validator also got stricter. Prompt and agent filenames are rejected when either side of the `__` separator is empty, and agent filenames are now checked for the namespace at all, which they were not before. Five tests cover those cases.

Two smaller items. The portability checklist in the review workflow now says all six harnesses and names Pi. The graceful degradation table now spells out what happens to a command body for Pi, which is that it is copied as written once Claude tool references are rewritten, `$ARGUMENTS` is left for Pi to substitute, and nothing is wrapped around it.

Two findings I did not take. The bare `dict` annotations in the Pi adapter match every other adapter in `tools/adapters/`, so changing only this one would make it the odd file out. The npm lockfile suggestion for the Pi CLI install is a structural change to how CI installs a smoke-test binary, so I left the direct version pin and added a comment noting that the transitive dependencies still resolve within their caret ranges.

Gates from the worktree root are green. Validate is strict clean across six harnesses, garden reports 0 errors and 18 warnings, lint passes including ty, and 661 tests pass.

https://claude.ai/code/session_01DZazzWVyb8MxPCuLC1w5Qo

